PR #1, opened 2020-04-06
[MOB-1326] add trackEvent

roninopf:


Iterable.ts
```diff
@@ -255,6 +255,11 @@ class Iterable {
         console.log("getInAppMessages");
         return RNIterableAPI.getInAppMessages().then((messages: Array<any>) => messages.map (message => {return IterableInAppMessage.fromDict(message)}))
     }
+
+    static track(event: String, dataFields: any | null) {
```
tapashmajumder, line 259:
```suggestion
    trackEvent(name: String, dataFields: any | null) {
```

Iterable.ts
```diff
@@ -255,6 +255,11 @@ class Iterable {
...
+        console.log("track")
```
tapashmajumder, line 260:
```suggestion
        console.log("trackEvent")
```

Iterable.ts
```diff
@@ -255,6 +255,11 @@ class Iterable {
...
+        return RNIterableAPI.track(event, dataFields)
```
tapashmajumder, line 261:
```suggestion
        return RNIterableAPI.trackEvent(event, dataFields)
```

index.d.ts
```diff
@@ -15,6 +15,7 @@ export class Iterable {
     static trackPushOpenWithCampaignId(campaignId: number, templateId: number, messageId: String | null, appAlreadyRunning: Boolean, dataFields: any | null): void
     static trackPurchase(total: number, items: Array<IterableCommerceItem>, dataFields: any | null): void
     static trackInAppOpen(message: IterableInAppMessage, location: IterableInAppLocation): void
+    static track(event: String, dataFields: any | null): void
```
tapashmajumder, line 18:
```suggestion
    static trackEvent(name: String, dataFields: any | null): void
```

ios/RNIterableAPI/RNIterableAPI.m
```diff
@@ -55,6 +55,7 @@ @interface RCT_EXTERN_REMAP_MODULE(RNIterableAPI, ReactIterableAPI, NSObject)
 RCT_EXTERN_METHOD(getInAppMessages: (RCTPromiseResolveBlock) resolve
                   rejecter: (RCTPromiseRejectBlock) reject)
 
-RCT_EXTERN_METHOD(track: (NSString *) event)
+RCT_EXTERN_METHOD(track: (NSString *) event
```
tapashmajumder, line 58:
```suggestion
RCT_EXTERN_METHOD(trackEvent: (NSString *) name
```

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -185,10 +185,10 @@ class ReactIterableAPI: RCTEventEmitter {
         resolver(IterableAPI.inAppManager.getMessages().map{ $0.toDict() })
     }
     
-    @objc(trackEvent:)
-    func track(event: String) {
+    @objc(trackEvent:dataFields:)
+    func track(event: String, dataFields: [AnyHashable: Any]?) {
```
tapashmajumder, line 189:
```suggestion
    func trackEvent(name: String, dataFields: [AnyHashable: Any]?) {
```

tapashmajumder:
Sorry that the description of this jira issue needs to be changed.

Since, unlike Obj C and Swift, Javascript goes only by the method name and not the param names, we should name the methods as `trackEvent` and not `track`.

I think in the future we should also look at renaming and deprecating our SDK. But that is for later.

Iterable.ts
tapashmajumder: it should be `static` trackEvent(name: etc.

PR #5, opened 2020-04-08
[MOB-1331] add handle universal link

roninopf:


Iterable.ts
```diff
@@ -315,6 +315,11 @@ class Iterable {
         console.log("updateEmail")
         RNIterableAPI.updateEmail(email)
     }
+
```
tapashmajumder, line 318: Please add comment for parameters of public method.

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -257,6 +261,12 @@ class ReactIterableAPI: RCTEventEmitter {
         IterableAPI.updateEmail(email, onSuccess: nil, onFailure: nil)
     }
     
+    @objc(handleUniversalLink:resolver:rejecter:)
+    func handleUniversalLink(link: String, resolver: RCTPromiseResolveBlock, rejecter: RCTPromiseRejectBlock) {
```
tapashmajumder, line 265: Since this is Swift we can name the method as below and this will not have any effect on javascript etc.
`handle(universalLink urlString: String, resolver...)`

tapashmajumder: This way we can all this method as `handle(universalLink: "https://somewhere.com/a/xyz" .... )`
Just a suggestion. I don't feel strongly about this. I think this would be more consistent with Swift naming.

tapashmajumder:
Sorry. It appears I may not have been very clear in my previous comment. We should only change the Swift definition to `handle(universalLink`. The Javascript and obj c should still refer to this as `handleUniversalLink`. This is because in Swift the method signature is `handle(universalLink:`). For Javascript the method signature is `handleUniversalLink()`. In Javascript there is no concept of method parameters being part of the signature, so the method name should be more descriptive whereas in Swift since the parameters are part of the method signature so the method name itself does not need to describe the parameters.

Iterable.ts
```diff
@@ -334,6 +334,15 @@ class Iterable {
...
+    /**
+     * 
+     * @param {String} universalLink URL in String form to be handled
+     */
+    static handle(universalLink: String): Promise<Boolean> {
```
tapashmajumder, line 342:
```suggestion
    static handleUniversalLink(universalLink: String): Promise<Boolean> {
```

index.d.ts
```diff
@@ -19,6 +19,7 @@ export class Iterable {
     static trackEvent(name: String, dataFields: any | null): void
     static updateUser(dataFields: any, mergeNestedObjects: Boolean): void
     static updateEmail(email: String): void
+    static handle(universalLink: String): Promise<Boolean>
```
tapashmajumder, line 22:
```suggestion
    static handleUniversalLink(universalLink: String): Promise<Boolean>
```

ios/RNIterableAPI/RNIterableAPI.m
```diff
@@ -73,6 +73,10 @@ @interface RCT_EXTERN_REMAP_MODULE(RNIterableAPI, ReactIterableAPI, NSObject)
 
 RCT_EXTERN_METHOD(updateEmail: (nonnull NSString *) email)
 
+RCT_EXTERN_METHOD(handle: (nonnull NSString *) universalLink
```
tapashmajumder, line 76:
```suggestion
RCT_EXTERN_METHOD(handleUniversalLink: (nonnull NSString *) universalLink
```

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -246,6 +250,12 @@ class ReactIterableAPI: RCTEventEmitter {
         IterableAPI.updateEmail(email, onSuccess: nil, onFailure: nil)
     }
     
+    @objc(handle:resolver:rejecter:)
```
tapashmajumder, line 253:
```suggestion
    @objc(handleUniversalLink:resolver:rejecter:)
```

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -246,6 +250,12 @@ class ReactIterableAPI: RCTEventEmitter {
...
+    func handle(universalLink: String, resolver: RCTPromiseResolveBlock, rejecter: RCTPromiseRejectBlock) {
+        ITBInfo()
+        resolver(IterableAPI.handle(universalLink: URL(string: universalLink)!))
```
tapashmajumder, line 256:
```suggestion
        if let url = URL(string: universalLink) {
            resolver(IterableAPI.handle(universalLink: url)
        } else {
           rejecter("", "invalid url", null)
        }
```

tapashmajumder:
My previous suggestions will be more clear when you look at how a method call will be made in Swift
`Iterable.handle(universalLink: "https://somewhere.com")`>
It is very clear from the method call what the `handle` method is doing because `universalLink` must be included in the method call.

The same method call in Javascript, if we do not make the changes I suggest will be
`Iterable.handle("https://somewhere.com")`. As you can see for Javascript, it is not clear at all what the `handle` method would do.

PR #12, opened 2020-05-05
[MOB-1487] - Support GetInAppMessages API

Ayyanchira:
Added support for getInAppMessages
Added Serialization methods to convert JSONObject and JSONArray to WritableMap and WritableArray as a react native bridge.
Added missing @ReactMethod annotations for previous commits
Modifying build gradle to latest Android SDK release


android/src/main/java/com/iterable/reactnative/Serialization.java
```diff
@@ -31,4 +50,154 @@ static IterableInAppDeleteActionType getIterableDeleteActionTypeFromInteger(@Nul
...
+        JSONArray inappMessagesJson = new JSONArray();
+
+        for (IterableInAppMessage message : inappMessages) {
+            JSONObject messageJson = new JSONObject();
+            try {
+                messageJson.putOpt("messageId", message.getMessageId());
+                //TODO: Incorporate public getter method for campaignId in next Android SDK release
+                //messageJson.putOpt("campaignId", message.getcampignId());
+                messageJson.putOpt("trigger", RNIterableInternal.getTriggerType(message));
+                messageJson.putOpt("createdAt", (int) message.getCreatedAt().getTime() / 1000);
```
Ayyanchira, line 66: Getting a wrong aura around this... 🙇‍♂️

vbabenkoru: I'd wrap this stuff into a method since you're doing the same a few lines later too.

Ayyanchira: Also type casting to (int) is changing the epoch date value. Keeping it `long` and assigning to JSON.

PR #13, opened 2020-05-08
[MOB-1486] - RN Android SDK - Support UpdateEmail

Ayyanchira:


android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -42,6 +42,12 @@ public void setEmail(String email) {
         IterableApi.getInstance().setEmail(email);
     }
 
+    @ReactMethod
+    public void updateEmail(String email) {
+        IterableLogger.d(TAG, "Update Email: " + email);
+        IterableApi.getInstance().updateEmail(email);
```
tapashmajumder, line 48: nit: `getEmail` method seems to delegate to `RNIterableInternal` class. Here it is not.

Ayyanchira: For methods which are package private and NOT public, we are using RNIterableInternal class to access the package private methods.
This is possible by having the RNIterableInternal class kept inside folder structure similar to that of Android’s IterableAPI class.

PR #56, opened 2020-07-13
[MOB-1858] - RN Android not able to pass null.

Ayyanchira:
Expecting developers to send -1(as mentioned in the docs) if the value is unknown. SDK will internally send null to Iterable's SDK


vbabenkoru:
Is this a limitation of RN that you can't have nullable arguments in native methods?

PR #57, opened 2020-07-14
[MOB-1861] - Add Firebase methods to wrapper method

Ayyanchira:
Developers would be able to call the two functions in their Firebase extended classes to achieve the same functionality as Native implementation.

Not sure though if this is an optimal solution.


vbabenkoru:
Isn't `IterableFirebaseMessagingService` accessible to developers directly?

Ayyanchira:
I believe just like all other public methods in Iterable's Android SDK, even the `IterableFirebaseMessagingService` is not accessible to end developers as well.

Ayyanchira:
Closing this ticket as we came up with a solution where `api` instead of `implementation` in gradle file can give the application direct access to IterableSDK

PR #59, opened 2020-07-21
[MOB-1893] - Access IterableSDK methods from RN Android

Ayyanchira:
Adding dependency using `api` keyword instead of `implementation` allows chained access to library. This allows app to use methods from internal library. More description given in the JIRA ticket with stack overflow link


Ayyanchira:
@bradumbaugh - This resolves our big question 😅 -? How would we access IterableSDK methods like the way iOS does???

bradumbaugh:
Nice, thanks @Ayyanchira !

PR #60, opened 2020-07-22
[MOB-1869] - Add support for LogLevel

Ayyanchira:


tapashmajumder:
Pushed a commit with some fixes to js code.

PR #69, opened 2020-08-07
[MOB-1966] - commenting out non released feature

Ayyanchira:


vbabenkoru:
Looks good, but I'd also check with @tapashmajumder and @roninopf.

PR #70, opened 2020-08-11
[MOB-1926] - RNAndroid - getHTMLContent

Ayyanchira:
Add a method getHtmlInAppContentForMessage for RN to be able to get messageContent

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -211,6 +211,26 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     }
 
     @ReactMethod
+    public void getHtmlInAppContentForMessage(String messageId, final Promise promise) {
+        IterableLogger.printInfo();
+        IterableInAppMessage message = RNIterableInternal.getMessageById(messageId);
+        if (message == null) {
+            promise.reject("", "Could not find message with id: " + messageId);
+            return;
+        }
+        JSONObject messageContent = Serialization.messageContentToJsonObject(message.getContent());
+        if (messageContent == null) {
+            promise.reject("", "messageContent is null for message id: " + messageId);
+            return;
+        }
+        try {
+            promise.resolve(Serialization.convertJsonToMap(messageContent));
+        } catch (JSONException e) {
+            promise.reject("", "Failed to convert JSONObject to ReadableMap");
+        }
+    }
+
+    @ReactMethod
     public void getAttributionInfo(Promise promise) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -210,6 +210,19 @@ public void execute(@Nullable Uri url) {
         });
     }
 
+    @ReactMethod
+    public void getHtmlInAppContentForMessage(String messageId, final Promise promise) {
+        IterableLogger.printInfo();
+        IterableInAppMessage message = RNIterableInternal.getMessageById(messageId);
+        if (message == null) {
```
tapashmajumder, line 217: Should reject promise here.

tapashmajumder: We should always either resolve or reject promise.

Ayyanchira: Yes.. I will look up in iOS counterpart what to send in reject in case of failure..

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -101,7 +101,7 @@ public void updateEmail(String email) {
     @ReactMethod
     public void updateEmail(String email, @Nullable String token) {
         IterableLogger.d(TAG, "updateEmail: " + email + ", token: " + token);
-        IterableApi.getInstance().updateEmail(email, token);
+//        IterableApi.getInstance().updateEmail(email, token);
```
tapashmajumder, line 104: Why is this commented out?

android/src/main/java/com/iterable/reactnative/Serialization.java
```diff
@@ -87,6 +87,26 @@ static CommerceItem commerceItemFromMap(JSONObject itemMap) throws JSONException
...
+        if (content == null) {
+            return null;
+        }
+
+        JSONObject messageContent = new JSONObject();
+        try {
+            messageContent.put("backgroundAlpha", content.backgroundAlpha);
+            messageContent.put("edgeInsets", content.padding);
+            messageContent.put("html", content.html);
+            if(messageContent.length() == 0) {
```
tapashmajumder, line 100: Nit: space after `if`

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
Ayyanchira: the authToken methods are private.. and not accessible from here. Have separate PR which comments the entire functions 😅

PR #72, opened 2020-08-28
[MOB-1969] - RN Android - Using rejecters where applicable

Ayyanchira:
Using rejecters where promises are available. In future, even in void functions, we should use rejecters to indicate failures in request fetched from RN side and promise resolve success if void methods are called successfully.

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -108,6 +108,11 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
 
     @ReactMethod
     public void getEmail(Promise promise) {
+        String email = RNIterableInternal.getEmail();
+        if (email == null){
+            promise.reject("", "email is null");
+            return;
+        }
         promise.resolve(RNIterableInternal.getEmail());
     }
 
@@ -142,6 +147,11 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
 
     @ReactMethod
     public void getUserId(Promise promise) {
+        String userId = RNIterableInternal.getEmail();
+        if (userId == null){
+            promise.reject("", "userId is null");
+            return;
+        }
         promise.resolve(RNIterableInternal.getUserId());
     }
 
@@ -203,6 +213,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     @ReactMethod
     public void showMessage(String messageId, boolean consume, final Promise promise) {
         if (messageId == null || messageId == "") {
+            promise.reject("", "messageId is null or empty");
             return;
         }
         IterableApi.getInstance().getInAppManager().showMessage(RNIterableInternal.getMessageById(messageId), consume, new IterableHelper.IterableUrlCallback() {
@@ -238,7 +249,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
         IterableLogger.printInfo();
         IterableAttributionInfo attributionInfo = IterableApi.getInstance().getAttributionInfo();
         if (attributionInfo == null) {
-            promise.resolve(null);
+            promise.reject("","No attribution info found");
             return;
         }
         try {
@@ -267,6 +278,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
             promise.resolve(Arguments.fromBundle(IterableApi.getInstance().getPayloadData()));
         } else {
             IterableLogger.d(TAG, "No payload data found");
+            promise.reject("","No payload data found");
         }
     }
 
@@ -345,6 +357,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
             promise.resolve(Serialization.convertJsonToArray(inAppMessageJsonArray));
         } catch (JSONException e) {
             IterableLogger.e(TAG, e.getLocalizedMessage());
+            promise.reject("","Failed to fetch messages with error " + e.getLocalizedMessage());
         }
     }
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -238,7 +249,7 @@ public void getAttributionInfo(Promise promise) {
         IterableLogger.printInfo();
         IterableAttributionInfo attributionInfo = IterableApi.getInstance().getAttributionInfo();
         if (attributionInfo == null) {
-            promise.resolve(null);
+            promise.reject("","No attribution info found");
```
vbabenkoru, line 252:
```suggestion
            promise.reject("", "No attribution info found");
```

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -267,6 +278,7 @@ public void getLastPushPayload(Promise promise) {
             promise.resolve(Arguments.fromBundle(IterableApi.getInstance().getPayloadData()));
         } else {
             IterableLogger.d(TAG, "No payload data found");
+            promise.reject("","No payload data found");
```
vbabenkoru, line 281:
```suggestion
            promise.reject("", "No payload data found");
```

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -345,6 +357,7 @@ public void getInAppMessages(Promise promise) {
             promise.resolve(Serialization.convertJsonToArray(inAppMessageJsonArray));
         } catch (JSONException e) {
             IterableLogger.e(TAG, e.getLocalizedMessage());
+            promise.reject("","Failed to fetch messages with error " + e.getLocalizedMessage());
```
vbabenkoru, line 360:
```suggestion
            promise.reject("", "Failed to fetch messages with error " + e.getLocalizedMessage());
```

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -108,6 +108,11 @@ public void updateEmail(String email) {
 
     @ReactMethod
     public void getEmail(Promise promise) {
+        String email = RNIterableInternal.getEmail();
+        if (email == null){
+            promise.reject("", "email is null");
```
vbabenkoru, line 113: Is this the expected behavior for a promise? I think `null` is a normal return value if the user hasn't logged in yet. I'm not sure about the semantics of a `Promise` but I thought it should only return a 'rejected' state if some error occurred?

Ayyanchira: We have to come to a consensus for that. This way RN developers dont have to rely on null checks for getEmail to proceed with their operations. And its like guarantee that something is always there if its a success.

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -108,6 +108,11 @@ public void updateEmail(String email) {
 
     @ReactMethod
     public void getEmail(Promise promise) {
+        String email = RNIterableInternal.getEmail();
+        if (email == null){
```
vbabenkoru, line 112:
```suggestion
        if (email == null) {
```

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
vbabenkoru: @tapashmajumder What do you think?

Ayyanchira: Had a chat with Tapash. Null email case will go to promise.resolve.

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -142,6 +142,11 @@ public void updateUser(ReadableMap dataFields, Boolean mergeNestedObjects) {
 
     @ReactMethod
     public void getUserId(Promise promise) {
+        String userId = RNIterableInternal.getEmail();
```
vbabenkoru, line 145: Doesn't the same apply to `getUserId` as `getEmail`?

Ayyanchira: Ah.. yes totally skipped it 😬.
Thanks Victor

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -238,7 +239,7 @@ public void getAttributionInfo(Promise promise) {
         IterableLogger.printInfo();
         IterableAttributionInfo attributionInfo = IterableApi.getInstance().getAttributionInfo();
         if (attributionInfo == null) {
-            promise.resolve(null);
+            promise.reject("", "No attribution info found");
```
vbabenkoru, line 242: Hmm, I didn't pay attention to this earlier, it seems similar to email and userId.. The value is `null`, so it shouldn't be an error.

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -267,6 +268,7 @@ public void getLastPushPayload(Promise promise) {
             promise.resolve(Arguments.fromBundle(IterableApi.getInstance().getPayloadData()));
         } else {
             IterableLogger.d(TAG, "No payload data found");
+            promise.reject("", "No payload data found");
```
vbabenkoru, line 271: Same as above.

PR #77, opened 2020-09-30
[MOB-2185] maintenance

roninopf:


Ayyanchira (approved):
Looks good. Tried running both iOS and Android and runs properly.. 👍

PR #79, opened 2020-10-01
Bug Fix: Android app not focused when using urlHandler or customActionHandler

### LINKED ISSUE #76: Android: App not focused on push notification press if backgrounded and custom url is handled
### Reproduction

1. Using Android app, use this lib and add a custom `urlHandler` which processes the `url` and returns `true`
1. Open the Android device, and then hit `home` button to place app in background (but still running)
1. Using Iterable Web UI, send a test push from a template with a `Push open action` of `Open URL`
1. On Android device, press the push

### Expected

1. App will process the push url, and come into focus

### Actual

1. App will process push url fully in background, but not come into focus

### Environment:

1. `@iterable/react-native-sdk` `1.0.12`
2. `react-native` `0.63.1`
3. `Android OS` `10`

### Suspected cause

Looking at [this code](https://github.com/Iterable/iterable-android-sdk/blob/817d006f7776a71f0b7beeef347e9dd666283dda/iterableapi/src/main/java/com/iterable/iterableapi/IterablePushActionReceiver.java#L85), it seems the Android SDK is designed to not focus the app if the event is handled. This may be fine for an Android SDK, but this RN SDK doesn't have a way to bring the app into focus if it reports the url is handled.

### Current workaround

For Android only, we return `false` in our `urlHandler`, basically disabling our custom url handler, and we make sure to use `urls` to double as valid deep links. This works for now, but is a somewhat limited solution.

jehartzog:
Fixes #76.

I went tracing through this and [iterable/iterable-android-sdk](https://github.com/Iterable/iterable-android-sdk) to figure out why we had some many issues with our Android app not bring brought into focus. I'll lay out what I found here, starting at the top level.

1. The actual `config.customActionHandler` you assign is fired [here](https://github.com/Iterable/react-native-sdk/blob/baea2e90c162d6bac72ab6a9f188f82cef5b5e8a/ts/Iterable.ts#L199), via `RNEventEmitter`. Note that nothing is actually done with the return value from your JS handler.

2. The event which executes the above is triggered in native Android [here](https://github.com/Iterable/react-native-sdk/blob/459201fa5da4d3022c69a9e7ad528d38ea63e581/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java#L382). The key issue here is **this function always returns `true`**, which tells the native code that any custom urls/actions are always handled. This also happens for [urls](https://github.com/Iterable/react-native-sdk/blob/459201fa5da4d3022c69a9e7ad528d38ea63e581/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java#L434).

3. The above function is called [here](https://github.com/Iterable/iterable-android-sdk/blob/817d006f7776a71f0b7beeef347e9dd666283dda/iterableapi/src/main/java/com/iterable/iterableapi/IterableActionRunner.java#L89), where if we have a `customActionHandler` defined, then we return the result of the above (which as we saw, is always `true`). Otherwise we return `false`.

4. The problem comes [here](https://github.com/Iterable/iterable-android-sdk/blob/85edd3322a442b733b5bc2a19a3f1b1aa2c57591/iterableapi/src/main/java/com/iterable/iterableapi/IterablePushActionReceiver.java#L83), where we execute the action, and if `handled === true` than we don't bring the app into focus. I believe this may have been doing sensibly for native `Android` apps, where they want to intercept a URL/action and can easily bring the app into foreground. However this is not the same for RN, as this code need to be run on native side.

I initially created a [fork of `iterable/iterable-android-sdk`](https://github.com/Iterable/iterable-android-sdk/compare/3.2.4...ourfabriq:fork/3.2.4-android-focus) which fixed the issue, but I later realize it would likely to be easier to fix for the RN use case, without breaking changes to the existing `iterable-android-sdk`, if we made the changes in this PR.

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -348,7 +348,8 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
         } catch (JSONException e) {
             IterableLogger.e(TAG, "Failed handling custom action");
         }
-        return true;
+        // The Android SDK will not bring the app into focus is this is `true`. It still respects the `openApp` bool flag.
+        return false;
     }
 
     @NonNull
@@ -386,7 +387,8 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
         } catch (JSONException e) {
             IterableLogger.e(TAG, e.getLocalizedMessage());
         }
-        return true;
+        // The Android SDK will not bring the app into focus is this is `true`. It still respects the `openApp` bool flag.
+        return false;
     }
 
     // ---------------------------------------------------------------------------------------


vbabenkoru (approved):
Thank you for the PR! Top notch investigative work @jehartzog! You're absolutely right: with RN, we can't assume the app calls `startActivity`, because RN apps are generally single-activity apps, and it's ok to send a launcher activity intent.
Thank you for the contribution. Very helpful.

roninopf:
Thank you @jehartzog! I've gone ahead and merged this, and it will be released soon.

PR #88, opened 2020-10-28
[MOB-2261] point to in-app animation bindings

roninopf:


vbabenkoru (approved):

ios/RNIterableAPI/Serialization.swift
```diff
@@ -143,8 +143,9 @@ extension IterableHtmlInAppContent {
         var dict = [AnyHashable: Any]()
         dict["type"] = type.rawValue
         dict["edgeInsets"] = edgeInsets.toDict()
-        dict["backgroundAlpha"] = backgroundAlpha
         dict["html"] = html
+        dict["shouldAnimate"] = shouldAnimate
+        dict["backgroundColor"] = backgroundColor ?? nil
```
Ayyanchira, line 148: Instead of backgroundColor, our actual JSON hierarchy looks like this
content
--inappDisplaySettings
----shouldAnimate
----bgColor
------"hex":String, 
------"alpha": Float 
--html

I remember the idea of having a seperate getHtmlContent method was to provide just the html cosmetic contents for those who need it.. I think contents inside inapp display settings are unnecessary for them. They can always have extra information in getInAppMessage where they can get everything about the IterableInAppMessage
@vbabenkoru need your thoughts on it as well.

Ayyanchira, line 148:
```suggestion
```

vbabenkoru: Looks like this is an interface between RN and native, so it doesn't necessarily have to follow the same structure as the JSON payload; do we expose these properties in native SDKs?

vbabenkoru: Yeah, basically if we don't deserialize or use these fields on Javascript side, we don't need to serialize them here.

vbabenkoru: And if we _want_ to make these fields available on Javascript side, we'll need to deseralize and expose them on JS side as well.

vbabenkoru (commented):
Looks more consistent now.
Do we want to expose these properties to RN though? Looks like they're public on iOS and Android.

vbabenkoru (approved):
Chatted during sync – looks like we don't want to make these new properties public yet, if there is demand, we'll add them.

PR #90, opened 2020-11-03
[MOB-2281] remove backgroundAlpha

roninopf:


Ayyanchira (approved):
Verified Android side ...runs properly... 👍

PR #106, opened 2021-02-01
[MOB-2459] in-app prioritization

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-2459](https://iterable.atlassian.net/browse/MOB-2459)

## ✏️ Description

* adds in-app prioritization variables to be read
* carries over in-app priority functionality from both beta SDKs

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,5 +22,5 @@ def getModuleVersion() {
 
 dependencies {
     implementation 'com.facebook.react:react-native:+'
-    api 'com.iterable:iterableapi:3.2.9'
+    api 'com.iterable:iterableapi:3.3.0-beta1'
 }
--- a/ts/IterableInAppClasses.ts
+++ b/ts/IterableInAppClasses.ts
@@ -155,6 +155,10 @@ class IterableInAppMessage {
    * Whether this inbox message has been read
    */
   readonly read: boolean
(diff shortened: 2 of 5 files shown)


ts/IterableInAppClasses.ts
```diff
@@ -174,6 +179,7 @@ class IterableInAppMessage {
     this.inboxMetadata = inboxMetadata
     this.customPayload = customPayload
     this.read = read
+    this.priorityLevel = priorityLevel
```
tapashmajumder, line 182: Adding this constructor param will cause compilation error in unit tests in `ts/__tests__/IterableInApp.spec.ts`.

I wonder why the tests are not failing and showing as passed.

tapashmajumder (approved):
+1

android/build.gradle
```diff
@@ -22,5 +22,5 @@ def getModuleVersion() {
 
 dependencies {
     implementation 'com.facebook.react:react-native:+'
-    api 'com.iterable:iterableapi:3.2.9'
+    api 'com.iterable:iterableapi:3.3.0-beta1'
```
Ayyanchira, line 25: Ah Jay! Just realized. It needs to have GitHub link instead of com.iterable link for the beta versions

Issue #92, opened 2020-11-18
Feature Recommendation: Move InApp Message 'read' data from local storage to API


jehartzog:
## TL/DR

The `read` field on all InApp messages is stored locally on a single device, so this field is reset if a user logs out, and then logs back in, or switches device. We had expected this to be stored on the API, to ensure a user doesn't seen old messages already seen, even if they switch devices or have to re-auth.

## Long version

We recently integrated the InApp messages to replace some of our custom built announcement features. Overall the integration went very smoothly, but there was one case where the iOS and Android sdks (and by extension, this one) worked counter to how we expected, and how the documentation implies it works.

Looking at this statement from the [docs](https://support.iterable.com/hc/en-us/articles/217517406-In-App-Messages-and-Mobile-inbox-):
> Iterable will not show the same in-app message to the same user more than once unless the message has been re-sent or selected in a mobile inbox.

There is a huge exception to this that wasn't clear until we completed our integration and did testing. 

When a message is `consumed`, the SDK's update the API ([android](https://github.com/Iterable/iterable-android-sdk/blob/f8a459722a688ee1cc5eb1bb4fddaf4f908b746f/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java#L256), [ios](https://github.com/Iterable/swift-sdk/blob/dd27e72e4198fbf8ea28aeed0cb5332eb121ffec/swift-sdk/Internal/InAppManager.swift#L486))

When a message is `read`, no API call is made ([android](https://github.com/Iterable/iterable-android-sdk/blob/f8a459722a688ee1cc5eb1bb4fddaf4f908b746f/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java#L133), [ios](https://github.com/Iterable/swift-sdk/blob/dd27e72e4198fbf8ea28aeed0cb5332eb121ffec/swift-sdk/Internal/InAppManager.swift#L142)).

[These docs](https://support.iterable.com/hc/en-us/articles/360018709631#displaying-in-app-messages) indicate that the only API calls made by the Iterable SDK when a Message is read is to track `In-App Open` event, which seems to indicate this was a deliberate decision to not create an API endpoint for storing this `read` data field.

## Effort on our use case

This means if we leave messages in the state of `not consumed` but `read` sitting in the message inbox, if the user re-auths, switches devices, then they will re-receive all those messages as they will all be immediately reset to `read = false`.

## Recommendation

[...]

PR #114, opened 2021-03-08
Add setReadForMessage to Android bridge

rayzeller:
## ✏️ Description

Add `setReadForMessage` to Android bridge

image

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -205,6 +205,11 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     }
 
     @ReactMethod
+    public void setReadForMessage(String messageId, boolean read) {
+        IterableApi.getInstance().getInAppManager().setRead(RNIterableInternal.getMessageById(messageId), read);
+    }
+
+    @ReactMethod
     public void getHtmlInAppContentForMessage(String messageId, final Promise promise) {
         IterableLogger.printInfo();
         IterableInAppMessage message = RNIterableInternal.getMessageById(messageId);


roninopf:
Hi @rayzeller, thanks for the PR, but we're fixing the message read state very soon. Was there a particular use for this you had that wasn't mentioned in this issue report? https://github.com/Iterable/react-native-sdk/issues/92

rayzeller:
I think when `read` is integrated into an API call, it will allow us to refactor our code so that we don't need to use the above method.

However, right now `setReadForMessage` is only defined for iOS devices. Trying to use it on Android would cause our devices to crash. I went ahead and PR-ed the patch we made so this wouldn't be the case. I guess we could also skip this API altogether with `set(message, 'read', true)` or a similar call, but right now the Typescript for `setReadForMessage` is misleading.

rayzeller:
Also @jehartzog and I work at the same company on a team of 3 engineers. Glad you're working on the issue ^_^. You have anything to add Eric?

jehartzog:
Nothing to add. Glad to hear there may be plans to transfer `read` field support to the API, that will enable the use cases we originally planned on. I agree w/ merging this PR in the interim, as using it currently will cause crashes on Android devices.

roninopf:
> I think when `read` is integrated into an API call, it will allow us to refactor our code so that we don't need to use the above method.
> 
> However, right now `setReadForMessage` is only defined for iOS devices. Trying to use it on Android would cause our devices to crash. I went ahead and PR-ed the patch we made so this wouldn't be the case. I guess we could also skip this API altogether with `set(message, 'read', true)` or a similar call, but right now the Typescript for `setReadForMessage` is misleading.

AH! Thank you. Another team member also informed me of this as well, I didn't realize this was happening since I usually work on the iOS bridge.

roninopf (approved):

PR #116, opened 2021-03-13
[MOB-2731] add case for auth token retrieval timing out

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-2731](https://iterable.atlassian.net/browse/MOB-2731)

## ✏️ Description

* add a catch for when the auth retrieval times out in the RN + iOS bridge

### DIFF
--- a/ios/RNIterableAPI/ReactIterableAPI.swift
+++ b/ios/RNIterableAPI/ReactIterableAPI.swift
@@ -573,7 +573,10 @@ extension ReactIterableAPI: IterableAuthDelegate {
                 }
             } else {
                 ITBInfo("authTokenRetrieval timed out")
-                completion(nil)
+                
+                DispatchQueue.main.async {
+                    completion(nil)
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -571,6 +571,9 @@ extension ReactIterableAPI: IterableAuthDelegate {
                 DispatchQueue.main.async {
                     completion(self.passedAuthToken)
                 }
+            } else {
+                ITBInfo("authTokenRetrieval timed out")
+                completion(nil)
```
vbabenkoru, line 576: The other `completion` call is made from the main queue, should we do the same here?

roninopf: AH yes it should be.

Ayyanchira (approved):

vbabenkoru (approved):

PR #131, opened 2021-04-27
[MOB-2862] - Remove deferred deeplinking.

tapashmajumder:
## 🔹 JIRA Ticket(s) if any

* [MOB-2862](https://iterable.atlassian.net/browse/MOB-2862)

## ✏️ Description

> Remove deferred deeplinking.


roninopf (approved):
+1. I tried to test these changes for compilation compatibility, but I think React Native broke. So, as long as you can make sure that's okay, it has my +1.

PR #153, opened 2021-07-21
[MOB-3159] split out in app stuff

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3159](https://iterable.atlassian.net/browse/MOB-3159)

## ✏️ Description

* refactors `IterableInAppMessage` and `IterableInAppManager` out to their own classes

### DIFF
--- a/ts/IterableInAppMessage.ts
+++ b/ts/IterableInAppMessage.ts
@@ -1,0 +1,124 @@
+'use strict'
+
+import { IterableUtil } from './IterableUtil'
+
+import { 
+  IterableInAppTrigger, 
+  IterableInAppTriggerType, 
+  IterableInboxMetadata,
+} from './IterableInAppClasses'
+
+/**
+ * Iterable in-app message
+ */
+ class IterableInAppMessage {
+    /**
+     * the ID for the in-app message
+     */
+    readonly messageId: string
+    /**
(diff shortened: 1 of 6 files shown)


ts/IterableInAppMessage.ts
```diff
@@ -0,0 +1,124 @@
...
+} from './IterableInAppClasses'
+
+/**
+ * Iterable in-app message
+ */
+ class IterableInAppMessage {
+    /**
+     * the ID for the in-app message
+     */
+    readonly messageId: string
```
martinmckenna, line 18: looks good to me. One question I was curious about last time I was looking over this. The [API docs](https://api.iterable.com/api/docs#In-app_getMessages) lists literally every single response property as possibly `undefined`. When I was building out the typings for these messages I just followed the API docs, but is it possible they're incorrect and some of this data is guaranteed to come back?

roninopf: Ah, no, the docs are correct, but we do processing on the returned payloads on the native side, so anything that gets to the React Native SDK by this point is either guaranteed data or filtered out beforehand. Same thing for the native SDKs' internal `getMessages` calls.

martinmckenna (approved):

PR #158, opened 2021-07-29
[MOB-3138] add message click functionality

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3138](https://iterable.atlassian.net/browse/MOB-3138)

## ✏️ Description

This pull request covers adding the click event to the messages cell. The message cell grays out and the status is updated to read if the message is unread. The number of unread messages will still need to be updated.


evangreer91:
New pull request will be created based on the new naming conventions.

PR #165, opened 2021-08-04
Mob 3138 add message click functionality using touchable

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3138](https://iterable.atlassian.net/browse/MOB-3138)

## ✏️ Description

This pull request adds click functionality to the react native SDK mobile inbox. This functionality has been tested using the react native dev app. In order to work with the current version of react native in the SDK, the touchableOpacity API was implemented. The newer pressable API implementation is created on a separate branch to be used using a more current version of react native.

### DIFF
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -7,18 +7,40 @@ import IterableInboxEmptyState from './IterableInboxEmptyState'
 import sampleMessages from './sampleMessageData.js'
 
 const IterableInbox = () => {
-   const message = "Inbox";
+   const inboxTitle = "Inbox";
+   const [selectedMessageIdx, setSelectedMessageIdx] = useState(null);
    const [messages, setMessages] = useState(sampleMessages);
 
+   function handleMessageSelect(index: number) {
+      let updatedMessages = messages;
+      let updatedMessage = messages[index];
+
+      if(!updatedMessage.read) {
+         updatedMessage.read = true
+      }
+
+      updatedMessages[index] = updatedMessage;
+      setMessages([...updatedMessages]);
+      setSelectedMessageIdx(index);
+   }   
+
    function showMessageList() {
-      return messages.length ? <IterableInboxMessageList messages={messages}></IterableInboxMessageList> : <IterableInboxEmptyState></IterableInboxEmptyState>
+      return (
+         <>
+            <Text style={styles.headline}>
+               {inboxTitle}
+            </Text>
+            {messages.length ? 
+               <IterableInboxMessageList 
+                  messages={messages}
+                  handleMessageSelect={handleMessageSelect}
+               ></IterableInboxMessageList> : 
+               <IterableInboxEmptyState></IterableInboxEmptyState>}
+         </>)
    }
 
    return(
       <View style={styles.container}>
-         <Text style={styles.headline}>
-            {message}
-         </Text>
          {showMessageList()}
       </View>
    )
--- a/ts/IterableInboxMessageCell.tsx
+++ b/ts/IterableInboxMessageCell.tsx
@@ -1,30 +1,76 @@
 "use strict"
-import React, { Component } from "react"
+import React, { Component, useState } from "react"
 import { 
    View, 
    Text, 
    StyleSheet, 
-   Pressable 
+   TouchableOpacity 
 } from "react-native"
 
 type MessageCellProps = {
+   index: number,
    message: {[key: string]: any},
+   handleMessageSelect: Function,
    last: boolean 
 }
 
-const IterableInboxMessageCell = ({ message, last }: MessageCellProps) => {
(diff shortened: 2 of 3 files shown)


ts/IterableInbox.tsx
```diff
@@ -7,18 +7,40 @@ import IterableInboxEmptyState from './IterableInboxEmptyState'
...
+      let updatedMessage = messages[index];
+
+      if(!updatedMessage.read) {
+         updatedMessage.read = true
+      }
+
+      updatedMessages[index] = updatedMessage;
+      setMessages([...updatedMessages]);
+      setSelectedMessageIdx(index);
+   }
```
notyourbear, line 25: small nit: 
personally I find it easier to follow what exactly the impact is of a function, when all the required components are passed in as part of the function's properties. given that you are relying on `messages` as an outside requirement, I find it easier to understand what's going on when the function contract looks like:

```
const handleMessageSelect = (index, messages) => {  
  ....code code code...
}
```
This has the added benefit that you don't really need to include the `let` statements since you're passing the messages in.


down on line `36` you can update the line to read: `handleMessageSelect={(index) => handleMessageSelect(index, messages)}`

what do you think?

ts/IterableInboxMessageCell.tsx
```diff
@@ -1,47 +1,111 @@
...
+   handleMessageSelect: Function,
    last: boolean 
 }
 
-const IterableInboxMessageCell = ({ message, last }: MessageCellProps) => {
+const IterableInboxMessageCell = ({ index, message, handleMessageSelect, last }: MessageCellProps) => {
    const unreadIndicator = "\u2022";
    const messageTitle = message.inboxMetadata.title
    const messageBody = message.inboxMetadata.subtitle
    const messageCreatedAt = message.createdAt
```
notyourbear, line 21: i see that our wonderful mobile team hates the ending `;`. i am sorry you have to deal with that. 😆 .

ts/IterableInboxMessageCell.tsx
```diff
@@ -1,47 +1,111 @@
...
 
+   const [active, setActive] = useState(false);
+
    function displayUnreadMessage() {
```
notyourbear, line 25: if i can make an architectural suggestion: i would put each of these individual react components into its own file and then import them all.  it's a big re-write, i realize, so i would _also_ suggest not doing that as part of this specific PR, but as a follow up ticket. Potentially even a tech-debt ticket to be picked up on test day. 

The reasons why this separation may be beneficial are:
1. the cognitive load to parse what's going on is much smaller when a file isn't needlessly long. 
2. from an organizational perspective, if in the future, multiple devs need to touch different parts of this file at the same time, there is less chance of merge conflict issues. 
3. (assuming that react-native works like react does) right now, every time that the parent component `IterableInboxMessageCell` is re-rendered because one of it's props changed, all of this code is recompiled again, needlessly.

ts/IterableInboxMessageCell.tsx
```diff
@@ -1,47 +1,111 @@
...
 
    return(
-      message.read ?
-         displayReadMessage() :
-         displayUnreadMessage()
+      <View>
+         {last ? 
+            (message.read ? displayReadLastMessage() : displayUnreadLastMessage()) :
+            (message.read ? displayReadMessage() : displayUnreadMessage())}
+      </View>
```
notyourbear, line 108: this is a little harder to follow than necessary, IMO. My general rule here is that: if whoever is on-call at 3am in the morning is just woken up and has to look at the code for the first time in their lives, can they easily parse what's happening?  I'd argue, a ternary inside of a ternary  is a bit complex. 

instead you could do something like: 

```
if (last) {
  return (
     <View>
       { message.read ? displayReadLastMessage() : displayUnreadLastMessage()) } 
     </View>
  )
} 

return (
      <View>
            { message.read ? displayReadMessage() : displayUnreadMessage())}
      </View>   
)
```

ts/IterableInbox.tsx
evangreer91: Good point, Patrick. I think it is better design to have the messages sent into the function as a parameter. I will make the change and send a new commit.

ts/IterableInboxMessageCell.tsx
evangreer91: Thanks. I will add the semi-colons.

ts/IterableInboxMessageCell.tsx
evangreer91: Thanks, Patrick. I will address this on another pull request.

ts/IterableInboxMessageCell.tsx
evangreer91: Yes, I will make the appropriate changes.

notyourbear (approved):
lgtm from a js perspective!

roninopf (approved):

PR #166, opened 2021-08-04
[MOB-3141] message display component and functionality

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3141](https://iterable.atlassian.net/browse/MOB-3141)

## ✏️ Description

This pull request includes the message display component and the associated functionality on the inbox component. This pull request uses the updated naming conventions and was tested in the react native dev app. The animations will still need to be implemented which will occur on another pull request

### DIFF
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -1,30 +1,30 @@
 'use strict';
-import React, { Component, useState } from 'react';
-import { Text, View, StyleSheet } from 'react-native';
-import IterableInboxMessageList from './IterableInboxMessageList';
-import IterableInboxEmptyState from './IterableInboxEmptyState';
-import IterableInboxMessageDisplay from './IterableInboxMessageDisplay';
-import sampleMessages from './sampleMessageData.js';
+import React, { Component, useState } from 'react'
+import { Text, View, StyleSheet } from 'react-native'
+import IterableInboxMessageList from './IterableInboxMessageList'
+import IterableInboxEmptyState from './IterableInboxEmptyState'
+import IterableInboxMessageDisplay from './IterableInboxMessageDisplay'
+import sampleMessages from './sampleMessageData.js'
 
 const IterableInbox = () => {
-   const inboxTitle = "Inbox";
-   const [isDisplayMessage, setIsDisplayMessage] = useState(false);
-   const [selectedMessageIdx, setSelectedMessageIdx] = useState(0);
-   const [messages, setMessages] = useState(sampleMessages);
+   const inboxTitle = "Inbox"
+   const [isDisplayMessage, setIsDisplayMessage] = useState(false)
+   const [selectedMessageIdx, setSelectedMessageIdx] = useState(0)
+   const [messages, setMessages] = useState(sampleMessages)
 
-   let selectedMessage = messages[selectedMessageIdx];
+   let selectedMessage = messages[selectedMessageIdx]
 
    function handleMessageSelect(index: number, messages: Array<any>) {
       if (!messages[index].read) {
          messages[index].read = true
       }
-      setMessages([...messages]);
-      setIsDisplayMessage(true);
-      setSelectedMessageIdx(index);
+      setMessages([...messages])
+      setIsDisplayMessage(true)
+      setSelectedMessageIdx(index)
    }
 
    function returnToInbox() {
-      setIsDisplayMessage(false);
+      setIsDisplayMessage(false)
    }
    
    function displayMessage(message: {[key: string]: any}) {
@@ -77,4 +77,4 @@ const styles = StyleSheet.create({
    }
 })
 
-export default IterableInbox;
+export default IterableInbox
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -27,7 +27,7 @@ const IterableInbox = () => {
       setIsDisplayMessage(false)
    }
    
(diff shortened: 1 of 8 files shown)


ts/IterableInboxMessageCell.tsx
```diff
@@ -1,47 +1,117 @@
...
+            onPress={() => {
+               setActive(!active);
+               handleMessageSelect(index);
+            }}
+         >
             <View style={styles.unreadIndicatorContainer}>
-               <Text style={styles.unreadIndicator}>{unreadIndicator}</Text>
+                  <View style={styles.unreadIndicator}/>
             </View>
             <View style={styles.unreadMessageContainer}>
```
roninopf, line 38: I think an abstraction/refactorization should be able to happen here, between a read message and an unread message, the only difference is the unread indicator, so, code wise, there shouldn't need to be duplication here.

The other thing to think about is also having the unread indicator have an on and off state so you can pass along just 1 message cell for either read/unread state.

ts/IterableInboxMessageList.tsx
```diff
@@ -1,17 +1,26 @@
...
+const IterableInboxMessageList = ({ messages, handleMessageSelect }: MessageListProps) => {
    function flaggedMessage(message: { [key: string]: any }, index: number) {
       return (index === messages.length - 1) ?
-         <IterableInboxMessageCell message={message} last={true} /> :
-         <IterableInboxMessageCell message={message} last={false} />
+         <IterableInboxMessageCell
+            index={index} 
+            message={message}
+            handleMessageSelect={(index: number) => handleMessageSelect(index, messages)} 
+            last={true} /> :
```
roninopf, line 18: A smaller version of the refactoring idea here; the only difference here is the `last` being `true` or `false`, otherwise the code is the same, but duplicated.

evangreer91: Yes I agree. It would be better to send down a flag that represents if the message is the last message on one message cell component. I will make the update.

ts/IterableInboxMessageCell.tsx
evangreer91: I agree. I will refactor the code to toggle the unread indicator in one display function.

evangreer91: I refactored the code to have a single display message function. But, it does implement the dreaded nested tertiary expression.

roninopf (approved):

tapashmajumder:
Please remove semi colons from every line. They are not needed in TypeScript.

ts/IterableInbox.tsx
```diff
@@ -1,25 +1,58 @@
...
-   const message = "Inbox";
+   const inboxTitle = "Inbox";
+   const [isDisplayMessage, setIsDisplayMessage] = useState(false);
+   const [selectedMessageIdx, setSelectedMessageIdx] = useState(0);
    const [messages, setMessages] = useState(sampleMessages);
 
+   let selectedMessage = messages[selectedMessageIdx];
+
+   function handleMessageSelect(index: number, messages: Array<any>) {
+      if(!messages[index].read) {
```
tapashmajumder, line 18:
```suggestion
      if (!messages[index].read) {
```

ts/IterableInbox.tsx
```diff
@@ -1,25 +1,58 @@
...
+      setMessages([...messages]);
+      setIsDisplayMessage(true);
+      setSelectedMessageIdx(index);
+   }
+
+   function returnToInbox() {
+      setIsDisplayMessage(false);
+   }
+   
+   function displayMessage(message: {[key: string]: any}) {
```
tapashmajumder, line 30: Also, I am not sure about the name 'displayMessage'. It appears that this message creates a 'IterableInboxMessageDisplay' element. Maybe 'createDisplayElement'? Just a suggestion.

ts/IterableInbox.tsx
```diff
@@ -1,25 +1,58 @@
...
+      return (
+         <IterableInboxMessageDisplay 
+            message={message}
+            returnToInbox={returnToInbox}
+         ></IterableInboxMessageDisplay>)
+   }
+
    function showMessageList() {
```
tapashmajumder, line 38: Same comment as `displayMessage' here.

ts/IterableInbox.tsx
evangreer91: I agree. Renaming to createMessageCell. In a future PR, I will separate this component into the message cell and the message cell click functionality.

ts/IterableInbox.tsx
evangreer91: I agree. I renamed the function to showMessageDisplay to align with showMessageList.

ts/IterableInbox.tsx
```diff
@@ -1,25 +1,58 @@
...
+   let selectedMessage = messages[selectedMessageIdx]
+
+   function handleMessageSelect(index: number, messages: Array<any>) {
+      if (!messages[index].read) {
+         messages[index].read = true
+      }
+      setMessages([...messages])
+      setIsDisplayMessage(true)
+      setSelectedMessageIdx(index)
+   }
```
notyourbear, line 24: so this is what i was talking about with the duplicate code from [the other pr](https://github.com/Iterable/react-native-sdk/pull/175/files), where you've actually made changes to how this will work. 

having this exact same code duplicated in two places means that I am no longer sure what I am reviewing. I guess, merge the other PR first so that those changes over there don't get overwritten.

PR #169, opened 2021-08-09
Added a dummy promise that returns true for android API initialization

### LINKED ISSUE #167: Android version of initializeWithApiKey isn't actually async
The type for the [initialize function](https://github.com/Iterable/react-native-sdk/blob/master/ts/Iterable.ts#L205) in typescript:
```
 static initialize(apiKey: string, config: IterableConfig = new IterableConfig()): Promise<boolean> {
 ```
is only accurate on iOS. On android there it is not resolving a promise so when we try to use the supposed promise returned from the function our app crashes. It works just fine on iOS though.

Android code function that is missing a promise:
https://github.com/Iterable/react-native-sdk/blob/master/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java#L72
```
    public void initializeWithApiKey(String apiKey, ReadableMap configReadableMap, String version) {
```

iOS function that does have it correctly:
https://github.com/Iterable/react-native-sdk/blob/master/ios/RNIterableAPI/ReactIterableAPI.swift#L54
```
    func initialize(apiKey: String,
                    config configDict: [AnyHashable: Any],
                    version: String,
                    resolver: @escaping RCTPromiseResolveBlock,
                    rejecter: @escaping RCTPromiseRejectBlock) {
```

We can work around this by just always assuming the initialize works on android and ignoring the promise, but it seems sketchy that the types aren't right and it would be nice to have some visibility if the initialization fails.

keeslinp:
## ✏️ Description

Fixes #167
This makes it match the iOS signature and fulfills what typescript says will be there.

This does not actually handle any errors so that should probably be added at some point in the future. This is mostly just to unblock ourselves as we are implementing the SDK. If someone wants to point me into the right direction for errors that I should be checking for then I am happy to do so. I see this as a small step in the right direction to prevent valid typescript applications from crashing unexpectedly.

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -69,7 +69,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     }
 
     @ReactMethod
-    public void initializeWithApiKey(String apiKey, ReadableMap configReadableMap, String version) {
+    public void initializeWithApiKey(String apiKey, ReadableMap configReadableMap, String version, Promise promise) {
         IterableLogger.d(TAG, "initializeWithApiKey: " + apiKey);
         IterableConfig.Builder configBuilder = Serialization.getConfigFromReadableMap(configReadableMap);
 
@@ -91,6 +91,9 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
 
         IterableApi.initialize(reactContext, apiKey, configBuilder.build());
         IterableApi.getInstance().setDeviceAttribute("reactNativeSDKVersion", version);
+        // TODO: Figure out what the error cases are and handle them appropriately
+        // This is just here to match the TS types and let the JS thread know when we are done initializing
+        promise.resolve(true);
     }
 
     @ReactMethod


Ayyanchira (approved):
Great suggestion @keeslinp! 👍

keeslinp:
@Ayyanchira do you know why the tests are failing? When I looked through the logs it seems that the tests are failing on iOS stuff which I haven't touched.

roninopf:
@keeslinp Thanks for the PR! For the tests, that was something on our side (we ended up switching to a slightly older version of RN in the integration test), which has since been merged into `master`. Feel free to pull in the latest, but we're okay with merging in this PR

keeslinp:
> @keeslinp Thanks for the PR! For the tests, that was something on our side (we ended up switching to a slightly older version of RN in the integration test), which has since been merged into `master`. Feel free to pull in the latest, but we're okay with merging in this PR

Who merges? I don't have access so if someone who has write access could do that, that would be super helpful.

roninopf:
> > @keeslinp Thanks for the PR! For the tests, that was something on our side (we ended up switching to a slightly older version of RN in the integration test), which has since been merged into `master`. Feel free to pull in the latest, but we're okay with merging in this PR
> 
> Who merges? I don't have access so if someone who has write access could do that, that would be super helpful.

I'll do it right now!

PR #174, opened 2021-08-12
[MOB - 3091] - Sync InApp on first launch

Ayyanchira:
## 🔹 JIRA Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-3091

## ✏️ Description

1. SDK Initialization now calls syncInApp internally.

This is because Android's InAppManager class is not attached to lifecycle of RN Android app at the very first launch and misses on calling syncInApp(), thus leading to failure in showing any new in-app messages.

Tested on Dev App
1. InApp sent in background while the app was killed is shown when the app starts up.
2. Backgrounding and Foregrounding the app does not call getMessages repeatedly as initialization is not called every time the app starts up. This means it becomes the job of developer to make sure initialization is not called repeatedly to avoid unnecessary getMessages call.

### DIFF
--- a/android/src/main/java/com/iterable/iterableapi/RNIterableInternal.java
+++ b/android/src/main/java/com/iterable/iterableapi/RNIterableInternal.java
@@ -36,4 +36,8 @@ public class RNIterableInternal {
         IterableApi.getInstance().setAttributionInfo(attributionInfo);
     }
 
+    public static void syncInApp() {
+        IterableApi.getInstance().getInAppManager().syncInApp();
+    }
+
 }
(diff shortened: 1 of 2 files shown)


tapashmajumder:
Why don't we call syncInApp automatically on initialize? Instead of a separate call.

Ayyanchira:
`syncInApp()` is neither called within initialization of `IterableApi` or `IterableInAppManager`. However, syncInApp() **gets called** right after IterableApi initialization using `onSwitchToForeground()` lifecycle callback in `IterableInAppManager`. This lifecycle methods are not registered and ready for RN during the very first launch.

Thats why an explicit syncInApp() call for RN setup.

Calling syncInApp() right inside IterableInAppManager initialization of Android SDK is also a good option. I am not sure if there was a reason behind keeping it that way.

tapashmajumder:
I think it would be a much better fix to call syncInApp when IterableInAppManager is started.
This way iOS and Android will be similar implementation.

Ayyanchira:
Made suggested change here https://github.com/Iterable/iterable-android-sdk/pull/371

tapashmajumder:
Can you check whether that fixes this issue? In that case, this PR may not be needed.

Ayyanchira:
Checked. It works 👍

tapashmajumder:
Awesome. You can close this PR then.

PR #175, opened 2021-08-12
[MOB-3260] separate message click functionality into its own component

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3260](https://iterable.atlassian.net/browse/MOB-3260)

## ✏️ Description

This pull request separates out the message click functionality into a separate component. The message list item will then be rendered by its own component. I thought this would be a good organization to help make testing and maintenance easier.


ts/IterableInbox.tsx
```diff
@@ -1,22 +1,39 @@
...
 
    function handleMessageSelect(index: number, messages: Array<any>) {
-      if(!messages[index].read) {
+      if (!messages[index].read) {
          messages[index].read = true
       }
-      setMessages([...messages]);
-      setSelectedMessageIdx(index);
-   }   
+      setMessages([...messages])
```
notyourbear, line 21: this all feels a little bit off to me. it may actually provided the intended result but it does so inspite of how react is built. 

when you are doing the following: 
```
if (!messages[index].read) {
    messages[index].read = true
}
```
you are actually making a mutable change to the messages object. which is why you then have to de-structure the `messages` array in the `setMessages` call for the state to get updated.  

as i said above, this all technically accomplishes what you're trying to do, but is a bit roundabout. 

instead, how do you feel about something like the following: 
```
function handleMessageSelect(index: number, messages: Array<any>) {
   const newMessages = messages.map((message, messageIndex) => {
      return (messageIndex === index) ?
         { ...message, read: true } : message
    }
   setMessages(newMessages);
   setIsDisplayMessage(true);
   setSelectedMessageIdx(index);
}
```

ts/IterableInboxClickableRow.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+const IterableInboxClickableRow = ({ index, message, handleMessageSelect, last }: MessageClickableRowProps) => {
+   const [active, setActive] = useState(false)
+
+   return(
+      <TouchableOpacity
+         style={
+            last
+               ? ((active) ? styles.pressedLastMessageCell : styles.lastMessageCell)
+               : ((active) ? styles.pressedMessageCell : styles.messageCell)
```
notyourbear, line 26: this is a personal nit, but I'd argue that a ternary inside of a ternary is hard to follow at first glance: 

how do you feel about creating a couple of variables  on line 20 like... 
```
const lastStyle = active ? styles.pressedLastMessageCell : styles.lastMessageCell;
const style = active ? styles.pressedMessageCell : styles.messageCell;
``` 

and then you can turn this into `style={ last ? lastStyle : style}`

ts/IterableInboxEmptyState.tsx
```diff
@@ -1,10 +1,10 @@
 "use strict";
-import React, { Component } from "react";
-import { View, Text, StyleSheet } from "react-native";
+import React, { Component } from "react"
+import { View, Text, StyleSheet } from "react-native"
 
 const IterableInboxEmptyState = () => {
-   const title = "No saved messages";
-   const subtitle = "Check again later!";
+   const title = "No saved messages"
```
notyourbear, line 6: semicolons are your friend 😆

ts/IterableInboxMessageList.tsx
```diff
@@ -1,31 +1,25 @@
...
-            index={index} 
-            message={message}
-            handleMessageSelect={(index: number) => handleMessageSelect(index, messages)}  
-            last={false} />
-   }
 
    function displayMessages() {
       return messages.map((message, index) => {
-         return flaggedMessage(message, index)
+         let last = (index === messages.length - 1) ? true : false
```
notyourbear, line 15: you don't need the `true : false` thing here.  `index === messages.length - 1` is either always `true` or `false`. 

`const isLast = index === messages.length - 1` will do.

ts/IterableInboxMessageList.tsx
```diff
@@ -1,31 +1,25 @@
...
+         return (         
+            <IterableInboxClickableRow
+               key={index}
```
notyourbear, line 18: does a message have an `id` property, or any other unique identifier? if possible, you'll want to rely on something more stable than the index of an array to act as your `key` -- this should be your last resort if there are no better identifiers.

ts/IterableInboxMessageListItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+import React, { Component, useState } from "react"
+import { 
+   View, 
+   Text, 
+   StyleSheet 
+} from "react-native"
+
+type MessageListItemProps = {
+   index: number,
+   message: {[key: string]: any},
```
notyourbear, line 11: you may want to create a specific type for a message rather than relying on `any`; this will help safe-guard you in the future.

ts/IterableInbox.tsx
evangreer91: I agree. This is a better approach. I should not be changing the state directly.

ts/IterableInboxEmptyState.tsx
evangreer91: Tapash told me to leave semicolons out since it is typescript

ts/IterableInboxClickableRow.tsx
evangreer91: I agree. The nested ternary did not look very good to me.

ts/IterableInboxMessageList.tsx
evangreer91: I will use the message ID for the key,

ts/IterableInboxMessageListItem.tsx
evangreer91: I agree. I created a message type file and imported the custom message type to all files that use the message type.

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -0,0 +1,61 @@
...
+const IterableInboxMessageDisplay = ({ message, returnToInbox }: MessageDisplayProps) => {
+   const messageTitle = message.inboxMetadata.title
+
+   return(
+      <View> 
+         <View style={styles.returnButtonContainer}>
+            <Icon 
+               name="ios-arrow-back" 
+               style={styles.returnButton} 
+               onPress={() => {returnToInbox()}}/>
```
notyourbear, line 21: `onPress={() => returnToInbox()}/>` makes this slightly more readable and causes no issues.

notyourbear (approved):
lgtm! great job!

ts/IterableInboxMessageDisplay.tsx
evangreer91: I will make the edit thanks.

PR #176, opened 2021-08-17
[MOB-3146] delete message functionality

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3146](https://iterable.atlassian.net/browse/MOB-3146)

## ✏️ Description

This pull request covers the message delete functionality. The message delete animation is encompassed in the IterableInboxSwipeableRow component. The deleteMessage function is passed down to this component. I had to do some modification to the message cell styling in the IterableInboxClickableRow component to match the height of the delete slider.

### DIFF
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -355,25 +355,18 @@ const IterableInbox = () => {
    const [selectedMessageIdx, setSelectedMessageIdx] = useState(0)
    const [messages, setMessages] = useState(sampleMessages)
 
-   // const selectedMessage = messages[selectedMessageIdx]
+   const selectedMessage = messages[selectedMessageIdx]
 
-   // function flagLastMessage(messages : Message[]) {
-   //    return messages.map((message : Message, index : number) => {
-   //       return (index === messages.length - 1) ? 
-   //          {...message, last: true} : {...message, last: false}
-   //    })
-   // }
+   function handleMessageSelect(index: number, messages: Message[]) {
+      const newMessages = messages.map((message : Message, messageIndex: number) => {
+         return (messageIndex === index) ?
+            {...message, read: true } : message
+      })
 
-   // function handleMessageSelect(index: number, messages: Message[]) {
-   //    const newMessages = messages.map((message : Message, messageIndex : number) => {
-   //       return (messageIndex === index) ?
-   //          {...message, read: true } : message
-   //    })
-
-   //    setMessages(newMessages)
-   //    setIsDisplayMessage(true)
-   //    setSelectedMessageIdx(index)
-   // }
+      setMessages(newMessages)
+      setIsDisplayMessage(true)
+      setSelectedMessageIdx(index)
+   }
 
    function deleteMessage(id: number, messages: Message[]) {
       let newMessages = sampleMessages.filter((message) => {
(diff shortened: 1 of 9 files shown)


ts/IterableInbox.tsx
```diff
@@ -1,30 +1,383 @@
...
-import { Text, View, StyleSheet } from 'react-native'
+import { Text, SafeAreaView, StyleSheet } from 'react-native'
+
 import IterableInboxMessageList from './IterableInboxMessageList'
 import IterableInboxEmptyState from './IterableInboxEmptyState'
 import IterableInboxMessageDisplay from './IterableInboxMessageDisplay'
-import sampleMessages from './sampleMessageData.js'
 import Message from './messageType'
 
+let sampleMessages = [
```
notyourbear, line 11: how do you feel about importing this mock data from another file. putting it here makes this file needlessly larger than it needs to be.

ts/IterableInboxSwipeableRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+   return(
+      <View>
+         <Animated.View style={deleteSlider}>
+            <Text style={textStyle}>DELETE</Text>   
+         </Animated.View>
+         <Animated.View 
+            style={[textContainer, position.getLayout()]}
+            {...panResponder.panHandlers}
+         >
+            <IterableInboxClickableRow
```
roninopf, line 123: I think in the future, it would be good to, instead, create an `IterableInboxRow`, and then give it attributes of being `Swipeable` and `Clickable`, rather than putting a `Clickable` row inside of a `Swipeable` row, it's more intuitive to understand it that way, and probably better, design wise.

ts/IterableInbox.tsx
```diff
@@ -1,30 +1,383 @@
...
    const [messages, setMessages] = useState(sampleMessages)
 
-   let selectedMessage = messages[selectedMessageIdx]
+   const selectedMessage = messages.find(message => message.messageId === selectedMessageId)
 
-   function handleMessageSelect(index: number, messages: Array<any>) {
-      const newMessages = messages.map((message, messageIndex) => {
-         return (messageIndex === index) ?
+   function handleMessageSelect(id: number, messages: Message[]) {
+      let newMessages = sampleMessages.map((message) => {
```
notyourbear, line 363: why is this line now using `sampleMessages` rather than the `messages` you're providing as a prop for the function?

roninopf (approved):

ts/IterableInbox.tsx
roninopf: Haha, I was gonna say, I'm not sure why Evan made this change, but I'll note that we'll be getting rid of this soon anyways, so take that as you will.

ts/IterableInbox.tsx
```diff
@@ -1,30 +1,383 @@
...
+         return (message.messageId === id) ?
             {...message, read: true } : message
       })
 
+      sampleMessages = newMessages
```
notyourbear, line 368: at no point should you need to reset `sampleMessages` either here nor in the `deleteMessages` function below. On line 358, you're setting the `messages` variable via `useState`. You shouldn't need to refer to `sampleMessages` at all after that point.

ts/IterableInboxClickableRow.tsx
```diff
@@ -1,38 +1,38 @@
...
 type MessageClickableRowProps = {
    index: number,
    message: Message,
    handleMessageSelect: Function,
-   last: boolean 
 }
 
-const IterableInboxClickableRow = ({ index, message, handleMessageSelect, last }: MessageClickableRowProps) => {
+const IterableInboxClickableRow = ({ 
+   index,
```
notyourbear, line 16: do you even still need this `index` prop passed down here? from what i can tell, neither this component, nor `IterableInboxMessageListItem` where it is passed down further make any use of it.

ts/IterableInboxSwipeableRow.tsx
evangreer91: Yes that is a good point. Let me see if I can reorganize in a way where the swipeable row and clickable row can be imported into the message list item.

ts/IterableInbox.tsx
evangreer91: I agree. The function should be pulling from the state. I made the update.

ts/IterableInboxClickableRow.tsx
evangreer91: The index prop is not needed. I will make the necessary update.

ts/IterableInbox.tsx
evangreer91: I do agree. But I am getting some strange behaviors from using the messages in state. Updating the sampleMessages solved that issue, but there should be a better solution.

PR #180, opened 2021-08-19
[MOB-3152] message list styling customization

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3152](https://iterable.atlassian.net/browse/MOB-3152)

## ✏️ Description

This pull request addresses the customization of the color scheme for the message row. This pull request also edits the rendering of the last message row so that the styling of the message rows are streamlined. The customizeStyles function updates any customization that is sent in as part of the messageRowCustomization object.

### DIFF
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -10,10 +10,16 @@ import sampleMessages from './sampleMessageData'
 import Message from './messageType'
 
 type inboxProps = {
-   customizations: {[key: string] : string}
+   navTitleCustomization: {[key: string] : string}
+   emptyStateCustomization: {[key: string] : string}
+   messageBodyCustomization: {[key: string] : any}
 }
 
-const IterableInbox = ({ customizations }: inboxProps) => {
+const IterableInbox = ({ 
+   navTitleCustomization,
+   emptyStateCustomization,
+   messageBodyCustomization 
+}: inboxProps) => {
    const defaultInboxTitle = "Inbox"
    const [isDisplayMessage, setIsDisplayMessage] = useState<boolean>(false)
    const [selectedMessageId, setSelectedMessageId] = useState<number>(1)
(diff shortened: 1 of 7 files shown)


ts/IterableInboxMessageListItem.tsx
```diff
@@ -11,27 +11,50 @@ import Message from './messageType'
...
+   customization: {[key: string]: any}
 }
 
-const IterableInboxMessageListItem = ({ message }: MessageListItemProps) => {
+const IterableInboxMessageListItem = ({ message, customization }: MessageListItemProps) => {
    const messageTitle = message.inboxMetadata.title
    const messageBody = message.inboxMetadata.subtitle
    const messageCreatedAt = message.createdAt
 
+   function customizeStyles(styles: {[key: string]: any}, customization: {[key: string]: any}) {
```
roninopf, line 22: I don't like this implementation as it has a lot of moving parts that are somewhat confusing to understand and it also seems like a roundabout way of implementing customization, but we can fix that in the next ticket. I was thinking that the customization props file just has a bunch of named keys/variables that the implementer would just create, and the rest of the code would just copy from, rather than doing something like this.

Also, we should separate the styling of the last row in a list from this. They are semantically separate things going on. And theoretically we should be able to integrate that code with either the main inbox rendering code or the message list code.

evangreer91: We can discuss this. My idea was that the customization would revert to a default styling if the customer does not provide a customization for that element. Putting it into one function that automates this seemed like a good idea.

ts/IterableInbox.tsx
```diff
@@ -8,16 +8,19 @@ import IterableInboxEmptyState from './IterableInboxEmptyState'
...
+import Customization from './customizationType'
 
 type inboxProps = {
-   customizations: {[key: string] : string}
+   customization: Customization
 }
 
-const IterableInbox = ({ customizations }: inboxProps) => {
+const IterableInbox = ({ 
+   customization
```
Ayyanchira, line 18: This is great! 👍

ts/customizationType.ts
```diff
@@ -0,0 +1,59 @@
+type Customization = {
```
Ayyanchira, line 1: So many customizable parameters 👍

ts/IterableInboxEmptyState.tsx
```diff
@@ -40,7 +43,7 @@ const styles = StyleSheet.create({
       paddingBottom: 25
    },
 
-   subtitle: {
+   body: {
       fontSize: 15,
       color: 'grey'
```
Ayyanchira, line 48: Wont this [body: fontSize] override with customizationType.ts -> [body: fontSize] ?

Ayyanchira: Oh got it. It rather passes it as argument to it.

roninopf (approved):

PR #185, opened 2021-08-24
[MOB - 3301] - urlHandler to return true

Ayyanchira:
## 🔹 JIRA Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-3301

## ✏️ Description

When a deeplink is clicked, IterableURLHandler in RN Bridge previously returned false regardless of whether the deeplink was handled or not handled.
This results in situation where deeplink will also be handled via client’s callback code, PLUS Iterable will try opening the link on browser.
To keep consistency with iOS, Android bridging code should also return true indicating that “deeplink urlcallback will be taken care by developers” and no need to further go and open links.

This also means that open URL functionality for React Native for deep links AND push notification wont work anymore.

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -453,8 +453,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
         } catch (JSONException e) {
             IterableLogger.e(TAG, e.getLocalizedMessage());
         }
-        // The Android SDK will not bring the app into focus is this is `true`. It still respects the `openApp` bool flag.
-        return false;
+        return true;
     }
 
     @Override


roninopf (approved):

vbabenkoru (changes requested):
Is there a reason why we were not waiting for the callback to return a value (like we do in the in-app handler) and were just returning `false`?

Ayyanchira:
Nice call actually... 👍 

Currently iOS is also returning hard coded false value without waiting for client's callback.

Will take a look 👍

Ayyanchira:
Returning `false` from bridging code to our SDK looks like the right approach. 

I just confirmed that urlHandler still opens links in the browser if urlHandler on client side(App.tsx) returns `false`. This is done by `Linking.openUrl()` function in our [Iterable.ts](https://github.com/Iterable/react-native-sdk/blob/a81afda72e722b87a054f2e9e578357c27d7437b/ts/Iterable.ts#L436-L440)

Having another callback loop like inAppHandler will become a duplicate behavior then.

I think I did not restart metro server when I was testing changing return values on App.tsx. Hot reload refreshes the app but old handlers were also working.

PR #187, opened 2021-08-24
[MOB-3296] incorporate inbox data model

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3294](https://iterable.atlassian.net/browse/MOB-3294)

## ✏️ Description

This pull request incorporates the data model into the react native inbox and allows for messages to be retrieved from the Iterable platform. The message list is populated with the user's messages and the message display shows the title.

### DIFF
--- a/ts/IterableInboxMessageListItem.tsx
+++ b/ts/IterableInboxMessageListItem.tsx
@@ -7,24 +7,53 @@ import {
    StyleSheet
 } from 'react-native'
 
-import Message from './messageType'
+import InboxRowViewModel from './InboxRowViewModel'
 import Customization from './customizationType'
 
 type MessageListItemProps = {
-   message: Message,
+   message: InboxRowViewModel,
    customization: Customization
 }
 
 const IterableInboxMessageListItem = ({ message, customization }: MessageListItemProps) => {
-   const messageTitle = message.inboxMetadata.title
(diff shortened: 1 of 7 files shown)


ts/IterableInboxMessageListItem.tsx
```diff
@@ -7,24 +7,53 @@ import {
...
    customization: Customization
 }
 
 const IterableInboxMessageListItem = ({ message, customization }: MessageListItemProps) => {
-   const messageTitle = message.inboxMetadata.title
-   const messageBody = message.inboxMetadata.subtitle
-   const messageCreatedAt = message.createdAt
+   const messageTitle = message.inAppMessage.inboxMetadata.title
+   const messageBody = message.inAppMessage.inboxMetadata.subtitle
+   const messageCreatedAt = convertTS(message.createdAt)
```
roninopf, line 21: Here, we should just pass along `new Date(message.createdAt)` rather than formatting it and doing all that, since the customer will specify how they will want that formatted.

evangreer91: Ok I got rid of the formatting. The message list item just renders the timestamp as is. I believe we want to do this instead of creating a Date object. The Date object will be used on future formatting functionality.

roninopf (approved):

PR #192, opened 2021-08-27
[MOB-3314] delete message from account

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3314](https://iterable.atlassian.net/browse/MOB-3314)

## ✏️ Description

This pull request provides the functionality to delete messages from the backend and sync with the user interface.

### DIFF
--- a/ts/InboxRowViewModel.ts
+++ b/ts/InboxRowViewModel.ts
@@ -8,6 +8,7 @@ type InboxRowViewModel = {
     imageUrl?: string
     createdAt?: Date
     read: boolean
+    last: boolean
     inAppMessage: IterableInAppMessage
 }
 
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -48,19 +48,19 @@ const IterableInbox = ({
          return (rowViewModel.inAppMessage.messageId === id) ?
             {...rowViewModel, read: true } : rowViewModel
       })
-      setRowViewModels(rowViewModels)
+      setRowViewModels(newRowViewModels)
       inboxDataModel.setItemAsRead(index)
 
       setIsDisplayMessage(true)
       setSelectedMessageIdx(index)
    }
 
-   const onDelete = (id: string, index: number, rowViewModels: InboxRowViewModel[]) => {useCallback(event => {
-      deleteMessage(id, index, rowViewModels)
-   }, rowViewModels)}
+   // const onDelete = (id: string, index: number, rowViewModels: InboxRowViewModel[]) => {useCallback(event => {
+   //    deleteMessage(id, index, rowViewModels)
+   // }, rowViewModels)}
 
    const deleteMessage = async (id: string, index: number, rowViewModels: InboxRowViewModel[]) => {
-      await inboxDataModel.deleteItem(index, IterableInAppDeleteSource.inboxSwipe)
+      await inboxDataModel.deleteItem(id, IterableInAppDeleteSource.inboxSwipe)
       fetchData()
       
       // let newRowViewModels = rowViewModels.filter((rowViewModel) => {
@@ -97,7 +97,7 @@ const IterableInbox = ({
                <IterableInboxMessageList 
                   rowViewModels={rowViewModels}
                   customizations={customizations}
-                  onDelete={(id: string, index: number) => onDelete(id, index, rowViewModels)}
+                  deleteMessage={(id: string, index: number) => deleteMessage(id, index, rowViewModels)}
                   handleMessageSelect={(id: string, index: number) => handleMessageSelect(id, index, rowViewModels)}
                />  : 
                <IterableInboxEmptyState customizations={customizations} />
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -8,6 +8,7 @@ import IterableInboxEmptyState from './IterableInboxEmptyState'
 import IterableInboxMessageDisplay from './IterableInboxMessageDisplay'
 
 import InboxRowViewModel from './InboxRowViewModel'
+import { IterableInAppDeleteSource } from './IterableInAppClasses'
 import IterableInboxDataModel from './IterableInboxDataModel'
 import Customization from './customizationType'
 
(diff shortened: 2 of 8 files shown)


ts/IterableInboxDataModel.ts
```diff
@@ -32,11 +32,11 @@ class IterableInboxDataModel {
         return this.inboxMessages
     }
 
-    deleteItem(row: number, deleteSource: IterableInAppDeleteSource) {
+    deleteItem(id: string, deleteSource: IterableInAppDeleteSource) {
```
roninopf, line 35: This is a good fix, but I think you should've created a new function instead of modifying `deleteItem` since it's tied into the data model's internal list of messages. The data model will need to be revised heavily (which I'll handle after this PR), and changing one of its current functions to work differently is confusing and not consistent with how the other functions (`getItemCount`, `getAlItems`, etc.) works.

For this PR, we'll have to leave it as is since it's a pretty interlocked big PR.

ts/IterableInboxMessageListItem.tsx
```diff
@@ -75,7 +75,7 @@ const defaultMessageListLayout = (message: InboxRowViewModel, customizations: Cu
       }
    })
 
-   styles = {...styles, ...customizations}
+   resolvedStyles = {...styles, ...customizations}
```
roninopf, line 78: I'm getting an error on this line when looking at this file, and I'm not sure why. We shouldn't check in code that has errors, much less an ambiguous error.

ts/IterableInboxMessageListItem.tsx
```diff
@@ -86,18 +86,18 @@ const defaultMessageListLayout = (message: InboxRowViewModel, customizations: Cu
       body,
       createdAt,
       messageRow
-   } = styles
+   } = resolvedStyles
 
-   function messageRowStyle(message: InboxRowViewModel) {
-      return message.last ? {...messageRow, borderBottomWidth: 1} : messageRow 
+   function messageRowStyle(rowViewModel: InboxRowViewModel) {
+      return rowViewModel.last ? {...messageRow, borderBottomWidth: 1} : messageRow
```
roninopf, line 92: As far as I can tell, this is the only use of `last` in the data model, I don't think it's a strong enough use case, especially since the way it's set and maintained isn't intuitive. My suggestion is let `IterableInboxMessageListItem` just take a singular `style` parameter, remove any notion of being a `last` row, and then just apply that style.

ts/InboxRowViewModel.ts
```diff
@@ -8,6 +8,7 @@ type InboxRowViewModel = {
     imageUrl?: string
     createdAt?: Date
     read: boolean
+    last: boolean
```
roninopf, line 11: I mentioned this in another comment, but I don't think this is a good design, because this type is representing a singular row, and `last` is something that is understood in an array/collection of rows. It's also not being maintained in a consistent, intuitive way; I think it could end up causing bugs later down the line.

ts/IterableInboxSwipeableRow.tsx
```diff
@@ -67,7 +69,7 @@ const IterableInboxSwipeableRow = ({
          toValue: {x, y: 0},
          duration: FORCING_DURATION,
          useNativeDriver: false   
-      })//.start(() => deleteMessage(message.inAppMessage.messageId))
+      }).start(() => deleteMessage(rowViewModel.inAppMessage.messageId, index))
```
roninopf, line 72: I couldn't find a good place to put this, but line 91 is also throwing an error for me that we should be able to resolve.

roninopf: Also noted here, we shouldn't need both the `messageId` and the `index` for deletion.

ts/IterableInboxMessageList.tsx
```diff
@@ -9,33 +9,33 @@ import InboxRowViewModel from './InboxRowViewModel'
...
-               key={message.inAppMessage.messageId}
+               key={rowViewModel.inAppMessage.messageId}
                index={index}
-               message={message}
+               rowViewModel={rowViewModel}
                messageListItemLayout={messageListItemLayout}
                customizations={customizations}
                //swipingCheck={(swiping : boolean) => setSwiping(swiping)}
-               //deleteMessage={(id: string) => deleteMessage(id)}
+               deleteMessage={(id: string, index: number) => deleteMessage(id, index)}
```
roninopf, line 38: We shouldn't need both the `messageId` and `index` here for deletion.

And we should also clarify what `id` it is.

ts/IterableInbox.tsx
```diff
@@ -1,13 +1,14 @@
 'use strict'
 
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useCallback } from 'react'
```
roninopf, line 3: We don't seem to be using `useCallback`, should this be removed?

ts/IterableInbox.tsx
```diff
@@ -22,46 +23,59 @@ const IterableInbox = ({
 }: inboxProps) => {
    const defaultInboxTitle = "Inbox"
    const [isDisplayMessage, setIsDisplayMessage] = useState<boolean>(false)
-   const [selectedMessageId, setSelectedMessageId] = useState<string>("")
-   const [messages, setMessages] = useState<InboxRowViewModel[]>([])
+   const [selectedMessageIdx, setSelectedMessageIdx] = useState<number>(0)
+   const [rowViewModels, setRowViewModels] = useState<InboxRowViewModel[]>([])
+
    const inboxDataModel = new IterableInboxDataModel()
+   // const inboxDeleteSource = new IterableInAppDeleteSource()
```
roninopf, line 30: We should delete irrelevant comments from the PR, they make reviewing more difficult than necessary.

ts/IterableInbox.tsx
```diff
@@ -22,46 +23,59 @@ const IterableInbox = ({
...
-   //    })
-   //    inboxDataModel.deleteItem(index, inboxSwipe)
-   //    //newMessages[newMessages.length - 1] = {...newMessages[newMessages.length - 1], last: true}
-   //    setMessages(newMessages)
-   // }
+   // const onDelete = (id: string, index: number, rowViewModels: InboxRowViewModel[]) => {useCallback(event => {
+   //    deleteMessage(id, index, rowViewModels)
+   // }, rowViewModels)}
+
+   const deleteMessage = async (id: string, index: number, rowViewModels: InboxRowViewModel[]) => {
```
roninopf, line 64: I'm not sure how necessary these comments around here are, but they're very confusing in here. They should either be removed, or another comment saying why they're here should be put in.

The `deleteMessage` function should probably be renamed to something else, it's not clear what a `Message` is. I would suggest `deleteRow` to keep the notion that this is an inbox with a bunch of rows.

Also, this function should not need both the `index` (which should be a `row` to prevent confusion), and `id` (which should be the `messageId` or 'inAppMessageId`), `index` isn't used anywhere in here, including in the commented code, and the rest of the components now have to utilize this signature.

ts/IterableInbox.tsx
```diff
@@ -81,13 +95,12 @@ const IterableInbox = ({
             <Text style={styles.headline}>
                {customizations.navTitle ? customizations.navTitle : defaultInboxTitle}
             </Text>
-            { messages.length ?
+            { rowViewModels.length ?
                <IterableInboxMessageList 
-                  messages={messages}
-                  messageListItemLayout={messageListItemLayout}
+                  rowViewModels={rowViewModels}
```
roninopf, line 100: It seems like `messageListItemLayout` got removed from here? I'm not sure why, but it's also showing as an error.

ts/IterableInbox.tsx
```diff
@@ -81,13 +95,12 @@ const IterableInbox = ({
...
                   customizations={customizations}
-                  //deleteMessage={(id: string) => deleteMessage(id, messages)}
-                  handleMessageSelect={(id: string, index: number) => handleMessageSelect(id, index, messages)}
+                  deleteMessage={(id: string, index: number) => deleteMessage(id, index, rowViewModels)}
```
roninopf, line 102: `inAppMessageId` and `index` etc. etc. you already know

ts/IterableInboxDataModel.ts
evangreer91: I will take a crack at creating a new function. We can discuss on Monday.

ts/IterableInboxMessageListItem.tsx
evangreer91: I addressed the error and added a const variable declaration.

ts/IterableInboxMessageListItem.tsx
evangreer91: I can pass down a prop that indicates the message list item is the last item in the list and use this to conditionally style the component. I will make the update.

ts/InboxRowViewModel.ts
evangreer91: Agreed. Removing from the InboxRowViewModel.

ts/IterableInboxSwipeableRow.tsx
evangreer91: I will resolve that error and get rid of index in the delete item call.

ts/IterableInboxMessageList.tsx
evangreer91: Agreed. I will clarify that it is the message id.

ts/IterableInbox.tsx
evangreer91: Agreed. I will remove useCallback.

ts/IterableInbox.tsx
evangreer91: Agreed. Deleted all unnecessary comments.

ts/IterableInbox.tsx
evangreer91: Renamed deleteMessage function to deleteRow. The id parameter is renamed to messageId.

ts/IterableInbox.tsx
evangreer91: Addressed.

ts/IterableInbox.tsx
evangreer91: Addressed.

roninopf (approved):

PR #204, opened 2021-09-08
[MOB-3298] inbox silent push integration

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3298](https://iterable.atlassian.net/browse/MOB-3298)

## ✏️ Description

* connects the `iterableInboxChanged` from the event to the RN inbox to both the iOS and Android bridges
* tested on iOS, but not tested on Android yet


evangreer91 (approved):

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -89,6 +89,8 @@ public void initializeWithApiKey(String apiKey, ReadableMap configReadableMap, S
             configBuilder.setAuthHandler(this);
         }
 
+        IterableApi.getInstance().getInAppManager().addListener(this);
+
         IterableApi.initialize(reactContext, apiKey, configBuilder.build());
```
Ayyanchira, line 94: add listener should happen after initialize(), which is the next line.

PR #205, opened 2021-09-17
[MOB-3391] update message display for rotation

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3391](https://iterable.atlassian.net/browse/MOB-3391)

## ✏️ Description

This pull request addresses the message list and message display conditional rendering on rotation of the device. The slide to delete functionality edits will be addressed on a separate PR.


ts/IterableInboxEmptyState.tsx
```diff
@@ -1,46 +1,74 @@
...
+
+   let {
+      container,
+      title,
+      body   
+   } = styles
+
+   container = {...container, height: height - navTitleHeight - tabBarHeight - tabBarPadding}
+
+   if(!isPortrait) {
```
roninopf, line 44: This got me thinking, is it really necessary to pass down `isPortrait` to everything? Shouldn't it be dynamically accounted for, rather than on render time?

ts/__tests__/IterableInboxMessageListItem.test.tsx
```diff
@@ -0,0 +1,115 @@
+import React from 'react'
+import { render } from '@testing-library/react-native'
```
roninopf, line 2: For some reason, I'm getting an error here when I run the tests locally

ts/useIsPortrait.tsx
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from "react"
+import { Dimensions } from "react-native"
```
roninopf, line 2: I saw in the docs that [useWindowDimensions](https://reactnative.dev/docs/usewindowdimensions) is preferred, and it seems like a good way for the code to rotate dynamically. Should we use that here?

ts/IterableInboxEmptyState.tsx
evangreer91: IsPortrait is accounted for dynamically. The useIsPortrait hook calculates if the orientation is portrait on orientation change.

ts/__tests__/IterableInboxMessageListItem.test.tsx
evangreer91: I will address the unit tests on a PR basis. This current unit test is not working.

ts/useIsPortrait.tsx
evangreer91: I will use useWindowDimensions in useIsPortrait hook. I will rename the useIsPortrait hook to useScreenDimensions hook. I will calculate and export the height and width within the useScreenDimensions hook.

ts/IterableInbox.tsx
```diff
@@ -187,24 +214,21 @@ const styles = StyleSheet.create({
 
    container: {
       flex: 1,
-      width: 2 * SCREEN_WIDTH,
       flexDirection: 'row',
-      height: '100%',
+      //height: '100%',
```
roninopf, line 218: Should we just remove this?

evangreer91: I will remove.

ts/IterableInbox.tsx
```diff
@@ -26,22 +24,42 @@ import IterableInboxCustomizations from './IterableInboxCustomizations'
...
+}: inboxProps) => {
    const defaultInboxTitle = "Inbox"
    const [selectedRowViewModelIdx, setSelectedRowViewModelIdx] = useState<number>(0)
    const [rowViewModels, setRowViewModels] = useState<InboxRowViewModel[]>([])
    const [loading, setLoading] = useState<boolean>(true)
    const inboxDataModel = new IterableInboxDataModel()
-
    const [animatedValue, setAnimatedValue] = useState<any>(new Animated.Value(0))
 
+   let { width, height, isPortrait } = useDeviceOrientation()
```
roninopf, line 49: Are these backwards? In `useDeviceOrientation`, it's `height` first.

evangreer91: Yes, good catch my bad. I will change that.

roninopf (approved):

Issue #198, opened 2021-09-02
Android crashes on trackEvent without dataFields


keeslinp:
`Iterable.trackEvent` does not require `dataFields` in the [typescript code](https://github.com/Iterable/react-native-sdk/blob/master/ts/Iterable.ts#L376). The [android SDK marks it as @Nullable](https://github.com/Iterable/iterable-android-sdk/blob/master/iterableapi/src/main/java/com/iterable/iterableapi/IterableApi.java#L490) as well, but the [serialization code](https://github.com/Iterable/react-native-sdk/blob/master/android/src/main/java/com/iterable/reactnative/Serialization.java#L267) doesn't check for it to be null so it crashes when it tries to access its member field.

You should probably update the annotations in the android code of this SDK and properly handle the null case. I'm getting around it by just passing in an empty object instead but it isn't a good first-impression to have crashes like that.

PR #207, opened 2021-09-21
[MOB-3362] create and use wrapped dataFields serialization

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3362](https://iterable.atlassian.net/browse/MOB-3362)

## ✏️ Description

* creates and wraps the `dataFields` serialization to allow `null`
* slight formatting consistency changes
* fixes https://github.com/Iterable/react-native-sdk/issues/198


Ayyanchira (approved):
Good one Jay! I was surprised to see that trackEvent with `null` was crashing before. Thanks for the fix.

PR #222, opened 2021-09-29
[MOB - 3467] - App link bug fix

Ayyanchira:
## 🔹 JIRA Ticket(s) if any

https://iterable.atlassian.net/browse/MOB-3467

## ✏️ Description

1. A wakeApp() method for RNAndroid which can invoke MainActivity
2. Before calling urlHandler, Iterable.ts will now wakeup the app if its minimized.

### DIFF
--- a/ts/Iterable.ts
+++ b/ts/Iterable.ts
@@ -1,6 +1,6 @@
 'use strict'
 
-import { NativeModules, NativeEventEmitter, Linking } from 'react-native'
+import { NativeModules, NativeEventEmitter, Linking, Platform } from 'react-native'
 import {
   IterableInAppShowResponse,
   IterableInAppLocation,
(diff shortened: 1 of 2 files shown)


ts/Iterable.ts
```diff
@@ -426,57 +439,70 @@ class Iterable {
...
         (dict) => {
           const url = dict["url"]
           const context = IterableActionContext.fromDict(dict["context"])
-          if (config.urlHandler!(url, context) == false) {
-            Linking.canOpenURL(url)
-              .then(canOpen => {
-                if (canOpen) { Linking.openURL(url) }
-              })
-              .catch(reason => { console.log("could not open url: " + reason) })
+          if (Platform.OS === "android") {
```
vbabenkoru, line 455: I would move this condition inside the `wakeApp` method. Because there is no iOS implementation, it will probably crash if called on iOS. The only thing preventing this is this condition here, and it is within a lot of other logic, so it can be easy. Also, someone else can add a call to `wakeApp` in some other method down the road, and that would also require a condition on `Platform.OS`. It is much easier to just have the check within the `wakeApp` method.

PR #229, opened 2021-10-05
[MOB-3507] render external links outside of webview component

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3507](https://iterable.atlassian.net/browse/MOB-3507)

## ✏️ Description

This pull request includes adding the event listener that renders external inks outside of the webview component in a separate window.


ts/IterableInboxMessageDisplay.tsx
```diff
@@ -83,6 +83,13 @@ const IterableInboxMessageDisplay = ({
                style={{ width: contentWidth }}
                onMessage={(event) => returnToInbox()}
                injectedJavaScript={JS}
+               onShouldStartLoadWithRequest={event => {
+                  if(event.url.slice(0,4) === 'http') {
```
Ayyanchira, line 87: We have a similar code used in Iterable.ts
https://github.com/Iterable/react-native-sdk/blob/a469215f694b4cf00ac150150fd1e888b0fe0f9e/ts/Iterable.ts#L436-L442

It will be better to use Linking.canOpenUrl() before making the Linking library to open url externally this way.

PR #241, opened 2021-10-14
[MOB-3460] date formatting

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3460](https://iterable.atlassian.net/browse/MOB-3460)

## ✏️ Description

* adds a way to format timestamps of inbox cells inside `IterableInboxDataModel`
* adds and uses a default formatting of timestamps

### DIFF
--- a/ts/IterableInboxMessageListItem.tsx
+++ b/ts/IterableInboxMessageListItem.tsx
@@ -10,10 +10,11 @@ import {
    StyleSheet
 } from 'react-native'
 
-import { InboxRowViewModel, IterableInboxCustomizations } from '.'
+import { InboxRowViewModel, IterableInboxCustomizations, IterableInboxDataModel } from '.'
 
 type MessageListItemProps = {
    last: boolean,
+   dataModel: IterableInboxDataModel,
    rowViewModel: InboxRowViewModel,
    customizations: IterableInboxCustomizations,
    messageListItemLayout: Function,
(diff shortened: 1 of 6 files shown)


ts/IterableInboxMessageListItem.tsx
```diff
@@ -139,15 +141,14 @@ const defaultMessageListLayout = (
...
    customizations,
    messageListItemLayout,
    isPortrait
 }: MessageListItemProps) => {
    return(
-      messageListItemLayout(last, rowViewModel) ?
-         messageListItemLayout(last, rowViewModel) :
-         defaultMessageListLayout(last, rowViewModel, customizations, isPortrait)  
+      messageListItemLayout(last, rowViewModel) ? messageListItemLayout(last, rowViewModel) : defaultMessageListLayout(last, dataModel, rowViewModel, customizations, isPortrait)  
    )
```
evangreer91, line 152: It seems like the data model is only used in the IterableInboxMessageListItem component for the formatted time stamp. Do we need to be importing the data model in the message list, swipeable row, and clickable row? To me it seems like it makes sense to only import it where it is used unless we anticipate it will be used in the other components in the future.

roninopf: You're right, but right now it's all structured in a nested manner. Because of this, I ended up creating a new ticket to create a new `IterableInboxCell` component that can intake all of those (and the data model), and also has the swiping,  clicking and item layout inside it, but that's a much bigger effort.

PR #252, opened 2021-10-29
[MOB-3639] integrating deep linking into webview

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3639](https://iterable.atlassian.net/browse/MOB-3639)

## ✏️ Description

This pull request updates the javascript executed by the webview component. The javascript now creates an event listener for each link in the HTML content. The link sends the url on click to the webview component. The handleHTMLMessage function acts based on the URL. If the URL is not "iterable://dismiss" or "iterable://delete", the urlHandler handles internal or external link. Logic is added to disable the opening of the HTML link if the link is a dismiss or delete link.


ts/IterableInboxMessageDisplay.tsx
```diff
@@ -46,13 +49,40 @@ const IterableInboxMessageDisplay = ({
...
+   let JS = `
+      const links = document.querySelectorAll('a')
+      links.forEach(link => {
+         link.addEventListener("click", () => {
+            window.ReactNativeWebView.postMessage(link.href)
+         })
+      })
+   `
+
+   let action = new IterableAction("", "", "")
```
Ayyanchira, line 61: Instead of passing empty strings, we wil have to create a IterableAction with based on what action we are going to perform:

Here is the link to Android SDK function of when we tap on a link on inapp.
https://github.com/Iterable/iterable-android-sdk/blob/3733809ee492f6fb45717a36df5ed6ec0e33a160/iterableapi/src/main/java/com/iterable/iterableapi/IterableInAppManager.java#L265

If it starts from iterable link, say action://, we define action type as action passed on to it(for e.g `delete`). Else, `openUrl` type.

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -46,13 +49,40 @@ const IterableInboxMessageDisplay = ({
...
 
+   const handleHTMLMessage = (event: any) => {
+      if(event.nativeEvent.data === 'iterable://delete') {
+         deleteRow(rowViewModel.inAppMessage.messageId)
+         returnToInbox()
+      } else if(event.nativeEvent.data === 'iterable://dismiss') {
+         returnToInbox()
+      } else {
+         if(Iterable.savedConfig.urlHandler) {
+            Iterable.savedConfig.urlHandler(event.nativeEvent.data, context)
```
Ayyanchira, line 80: This takes care of the deep link where urlHandler is called. But it still wont takes care of scenario where they return false and SDK opens the url in browser.

evangreer91: Yes, I plan to add the external linking on a separate PR today.

Ayyanchira: sounds good 👌

ts/IterableInboxMessageDisplay.tsx
evangreer91: Per our huddle, I added a type and the url passed to the urlHandler as data to the IterableAction.

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -7,25 +7,29 @@ import {
   ScrollView,  
   StyleSheet,
   Platform,
+  Alert,
```
Ayyanchira, line 10: I dont see Alert being used. You can remove unused imports.

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -46,13 +50,59 @@ const IterableInboxMessageDisplay = ({
...
+         if(link.href === "iterable://dismiss") {
+            link.class = "dismiss"   
+         }
+
+         if(link.href === "iterable://delete") {
+            link.class = "delete"
+         }
+
+         link.addEventListener("click", () => {
+            if(link.class === "dismiss") {
```
Ayyanchira, line 65: I couldn't figure out why was it necessary to have link.class = "dismiss" or "delete" when we could just check it in if condition here? 
Is link.class being used else where?

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -46,13 +50,59 @@ const IterableInboxMessageDisplay = ({
...
+               window.ReactNativeWebView.postMessage("iterable://dismiss")
+            } else if(link.class === "delete") {
+               window.ReactNativeWebView.postMessage("iterable://delete")
+            } else {
+               window.ReactNativeWebView.postMessage(link.href)
+            }
+         })
+
+         if(link.href === "iterable://dismiss" || link.href === "iterable://delete") {
+            link.href = "javascript:void(0)"
```
Ayyanchira, line 75: This is to avoid accidental multiple link clicks, right? Or is there usage too?

Ayyanchira (changes requested):

ts/IterableInboxMessageDisplay.tsx
evangreer91: Yes. Alert was being used for debugging. I will remove.

ts/IterableInboxMessageDisplay.tsx
evangreer91: I added the class label since I have to change the href to deactivate the HTML link. Without the deactivation, the HTML link tries to open the iterable action. By changing the href to "javascript: void(0)", the link will redirect to stay on the same page. On a future PR, I will be working on the more generalized case by incorporating internal links into the logic.

ts/IterableInboxMessageDisplay.tsx
evangreer91: As addressed above, changing the href of the link prevents the HTML from trying to open the iterable action.

Ayyanchira (approved):

roninopf (approved):

PR #253, opened 2021-11-01
[MOB-3661] add external linking to webview component

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3661](https://iterable.atlassian.net/browse/MOB-3661)

## ✏️ Description

This pull request includes adding the external linking logic outside of the handleHTMLmessage function. The external linking needs to be dealt with by the onShouldStartLoadingWithRequest prop since we need to tell the webview to stop loading if the link goes to an external link (starts with http).


ts/IterableInboxMessageDisplay.tsx
```diff
@@ -70,17 +71,27 @@ const IterableInboxMessageDisplay = ({
    })
 
    const handleHTMLMessage = (event: any) => {
-      if(event.nativeEvent.data === 'iterable://delete') {
+      let url = event.nativeEvent.data
+      if(url === 'iterable://delete') {
```
roninopf, line 75: Not needed for this PR but we should try to be more consistent with syntax, like spacing between an `if` and the condition

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -70,17 +71,27 @@ const IterableInboxMessageDisplay = ({
...
-         if(Iterable.savedConfig.urlHandler) {
-            Iterable.savedConfig.urlHandler(event.nativeEvent.data, context)
+         if(Iterable.savedConfig.urlHandler && Iterable.savedConfig.urlHandler(url, context)) {
+            Iterable.savedConfig.urlHandler(url, context)
+            returnToInbox()
          }
+      }
+   }
+
+   const openExternalURL = (event: any) => {
```
roninopf, line 88: Also not needed for this PR, but I think we should try to use `function` more, I think the advantages are helpful for a method like this, like in [this article](https://softwareengineering.stackexchange.com/questions/364086/why-use-const-foo-instead-of-function-foo).

Ayyanchira: My vote goes for function too 👍

ts/IterableInboxMessageDisplay.tsx
Ayyanchira: Tip: If you are using VS Code, select the code section you edited and use `Opt + Shift + F`. This will indent the code

evangreer91: I went ahead and formatted all the inbox files

ts/IterableInboxMessageDisplay.tsx
evangreer91: I will use function going forward.

PR #256, opened 2021-11-02
[MOB-3384] on screen cell tracking

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3384](https://iterable.atlassian.net/browse/MOB-3384)

## ✏️ Description

* tracks which messages in the inbox are showing (100% visibility and present for at least 0.5s)
* proper session start and end tracking will be handled in other tickets
* to see the row tracking functioning (within RN), you can uncomment the logging functions in `IterableInboxMessageList.tsx` and watch the RN console print out what rows are visible


evangreer91 (approved):

ts/IterableInboxMessageList.tsx
```diff
@@ -52,18 +55,56 @@ const IterableInboxMessageList = ({
       )
    }
 
+   function getInAppMessageFromViewToken(viewToken: ViewToken): IterableInAppMessage {
```
Ayyanchira, line 58: What is viewToken?

ts/IterableInAppMessage.ts
```diff
@@ -84,6 +84,21 @@ class IterableInAppMessage {
     this.priorityLevel = priorityLevel
   }
 
+  static fromInApp(inAppMessage: IterableInAppMessage) {
+    return new IterableInAppMessage(
```
Ayyanchira, line 88: Not sure but looks strange that the class can accept the object of same type and return a new class out of it.. 😐 might need Patrick or Martys opinion to see if it is the best way.

ts/IterableInboxMessageList.tsx
```diff
@@ -52,18 +55,56 @@ const IterableInboxMessageList = ({
...
+      return IterableInAppMessage.fromInApp(viewToken.item["inAppMessage"] as IterableInAppMessage)
+   }
+
+   // function getInAppMessagesFromViewTokens(viewTokens: Array<ViewToken>): Array<IterableInAppMessage> {
+   //    return viewTokens.map(getInAppMessageFromViewToken)
+   // }
+
+   function getRowInfosFromViewTokens(viewTokens: Array<ViewToken>): Array<InboxImpressionRowInfo> {
+      return viewTokens.map(
+         function(viewToken) {
```
Ayyanchira, line 68: Can this function be moved outside? If inline, I think keyword 'function' wont be needed

Ayyanchira: May be its just cosmetic change

ts/IterableInboxMessageList.tsx
```diff
@@ -52,18 +55,56 @@ const IterableInboxMessageList = ({
...
+      )
+   }
+
    const inboxSessionViewabilityConfig: ViewabilityConfig = {
-      minimumViewTime: 0.5,
-      itemVisiblePercentThreshold: 50,
-      waitForInteraction: true
+      minimumViewTime: 500,
+      itemVisiblePercentThreshold: 100,
+      waitForInteraction: false
```
Ayyanchira, line 84: 👍Liked that its configurable

Ayyanchira (commented):
Left some comments

ts/IterableInboxMessageList.tsx
roninopf: It's related to the RN FlatList component, it contains info about specific cells when tracking. [API docs](https://reactnative.dev/docs/0.65/viewtoken)

ts/IterableInAppMessage.ts
roninopf: There's no way to create more constructors, and for some reason, this is necessary to avoid errors within TypeScript. Will ask a TS expert.

ts/IterableInboxMessageList.tsx
roninopf: I'd prefer not to move it outside for now because it's pretty parasitic (only works within this component).

I also can't remove `function`, it causes errors.

ts/IterableInAppMessage.ts
roninopf: After talking with Marty and Patrick about this, the OVERALL way I'm doing this is probably the best way to handle it due to it being a weird API, but a suggested change would be to have `IterableInAppMessage.ts` take a `ViewToken` rather than an `IterableInAppMessage` here.

I'll do this as a separate ticket since it's only relevant to this specific feature and functionally doesn't change anything.

Ayyanchira: sounds good 👍

Ayyanchira (approved):

PR #257, opened 2021-11-02
[MOB-3622] track in-app close events

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3622](https://iterable.atlassian.net/browse/MOB-3622)

## ✏️ Description

This pull request includes adding in-app close events to the handleHTMLmessage function when the dismiss link is clicked and to the onPress prop of the return to inbox button.

### DIFF
--- a/ts/IterableInboxMessageDisplay.tsx
+++ b/ts/IterableInboxMessageDisplay.tsx
@@ -81,7 +81,7 @@ const IterableInboxMessageDisplay = ({
          returnToInbox()
       } else if(url === 'iterable://dismiss') {
          returnToInbox()
-         Iterable.trackInAppClose(rowViewModel.inAppMessage, IterableInAppLocation.inbox, IterableInAppCloseSource.link, undefined)
+         Iterable.trackInAppClose(rowViewModel.inAppMessage, IterableInAppLocation.inbox, IterableInAppCloseSource.link, null)
       } else {
          if(Iterable.savedConfig.urlHandler && Iterable.savedConfig.urlHandler(url, context)) {
             Iterable.savedConfig.urlHandler(url, context)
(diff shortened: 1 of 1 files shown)


ts/IterableInboxMessageDisplay.tsx
```diff
@@ -74,13 +74,14 @@ const IterableInboxMessageDisplay = ({
...
 
-      Iterable.trackInAppClick(rowViewModel.inAppMessage, IterableInAppLocation.inApp, url)
+      Iterable.trackInAppClick(rowViewModel.inAppMessage, IterableInAppLocation.inbox, url)
       
       if(url === 'iterable://delete') {
          deleteRow(rowViewModel.inAppMessage.messageId)
          returnToInbox()
       } else if(url === 'iterable://dismiss') {
          returnToInbox()
+         Iterable.trackInAppClose(rowViewModel.inAppMessage, IterableInAppLocation.inbox, IterableInAppCloseSource.link, 'iterable://dismiss')
```
roninopf, line 84: Whoops. I think our `trackInAppClose` isn't complete, the `clickedUrl` field here should just be `null`, so we should build that support in.

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -101,7 +102,11 @@ const IterableInboxMessageDisplay = ({
    return(
       <View style={updatedMessageDisplayContainer}>
          <View style={returnButtonContainer}>
-            <TouchableWithoutFeedback onPress={() => returnToInbox()}>
+            <TouchableWithoutFeedback 
+               onPress={() => {
+                  returnToInbox()
+                  Iterable.trackInAppClose(rowViewModel.inAppMessage, IterableInAppLocation.inbox, IterableInAppCloseSource.back, 'iterable://dismiss')
+               }}>
```
roninopf, line 109: Same to here, we should allow `clickedUrl` to be `null`.

ts/IterableInboxMessageDisplay.tsx
evangreer91: I will make the update.

ts/IterableInboxMessageDisplay.tsx
evangreer91: I will make the update.

ts/IterableInboxMessageDisplay.tsx
evangreer91: I added undefined for the clickedUrl field

ts/IterableInboxMessageDisplay.tsx
evangreer91: I added undefined for the clickedUrl field.

roninopf (approved):

PR #258, opened 2021-11-03
[MOB-3672] allow track in app close to take null clickedUrl

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3672](https://iterable.atlassian.net/browse/MOB-3672)

## ✏️ Description

* allows `trackInAppClose` to take a `null` `clickedUrl` parameter when there's no associated URL with the event (e.g. back button or `dismiss`)
* slight refactoring to use standardized API calls

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -290,10 +290,12 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     @ReactMethod
     public void trackInAppOpen(String messageId, @Nullable Integer location) {
         IterableInAppMessage message = RNIterableInternal.getMessageById(messageId);
+
         if (message == null) {
-            IterableLogger.d(TAG, "Failed to get InApp for message id : " + messageId);
+            IterableLogger.d(TAG, "Failed to get in-app for message ID: " + messageId);
             return;
         }
+
         IterableApi.getInstance().trackInAppOpen(message, Serialization.getIterableInAppLocationFromInteger(location));
     }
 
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -290,41 +290,59 @@ public void handleAppLink(String uri, Promise promise) {
...
+
+        if (inAppCloseLocation == null) {
+            IterableLogger.d(TAG, "in-app close location is null");
+            return;
+        }
+
+        if (closeAction == null) {
+            IterableLogger.d(TAG, "in-app close action is null");
+        }
+
```
Ayyanchira, line 344: may be we can also log clickedUrl is null just like the others and continue passing null

Ayyanchira (approved):
Added a nit-picking comment. 
Looks good.

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
roninopf: Whoops, there's supposed to be a `return;` there, it's not just a logging thing. Good catch

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -290,41 +290,60 @@ public void handleAppLink(String uri, Promise promise) {
...
         if (message == null) {
-            IterableLogger.d(TAG, "Failed to get InApp for message id : " + messageId);
+            IterableLogger.d(TAG, "Failed to get in-app for message ID: " + messageId);
             return;
         }
+
         if (clickedUrl == null) {
-            IterableLogger.d(TAG, "clickedURL is null");
+            IterableLogger.d(TAG, "clickedUrl is null");
```
Ayyanchira, line 313: Here we are return and stop the process if clickedUrl is null. Should we not pass and let it stay null?

Ayyanchira (approved):
Added another comment. Rest looks good to me :)

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
roninopf: That's specific to `trackInAppClick`, which needs it for the event (otherwise the event doesn't make sense)

PR #263, opened 2021-11-08
[MOB-3620&3622] in-app click and close events for HTML links

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3620](https://iterable.atlassian.net/browse/MOB-3620)
* [MOB-3622](https://iterable.atlassian.net/browse/MOB-3622)

## ✏️ Description

This pull request includes the click and close events for the HTML links. This pull request also includes the logic for suppressing the HTML links.

### DIFF
--- a/ts/IterableInboxMessageDisplay.tsx
+++ b/ts/IterableInboxMessageDisplay.tsx
@@ -13,8 +13,18 @@ import {
 import { WebView } from 'react-native-webview'
 import Icon from 'react-native-vector-icons/Ionicons'
 
-import { InboxRowViewModel, IterableHtmlInAppContent, IterableEdgeInsets } from '.'
-import { Iterable, IterableAction, IterableActionSource, IterableActionContext } from './Iterable'
+import { 
+   InboxRowViewModel, 
+   IterableHtmlInAppContent, 
+   IterableEdgeInsets,
+   IterableInAppLocation,
+   IterableInAppCloseSource,
+   Iterable,
+   IterableAction,
+   IterableActionContext 
+} from '.'
+
+import { IterableActionSource } from './Iterable'
 
 type MessageDisplayProps = {
    rowViewModel: InboxRowViewModel,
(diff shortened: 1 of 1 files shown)


ts/IterableInboxMessageDisplay.tsx
```diff
@@ -86,20 +96,24 @@ const IterableInboxMessageDisplay = ({
...
+      Iterable.trackInAppClick(rowViewModel.inAppMessage, IterableInAppLocation.inbox, URL)
+
       if(URL === 'iterable://delete') {
          deleteRow(rowViewModel.inAppMessage.messageId)
          returnToInbox()
       } else if(URL === 'iterable://dismiss') {
          returnToInbox()
+         Iterable.trackInAppClose(rowViewModel.inAppMessage, IterableInAppLocation.inbox, IterableInAppCloseSource.link)
       } else {
          if(Iterable.savedConfig.urlHandler) {
```
Ayyanchira, line 109: Looks like you can combine else if { } here

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -86,20 +96,24 @@ const IterableInboxMessageDisplay = ({
...
             let action = new IterableAction("openUrl", URL, "")
             let source = IterableActionSource.inApp
             let context = new IterableActionContext(action, source)
 
             Iterable.savedConfig.urlHandler(event.nativeEvent.data, context)
+            returnToInbox()
```
Ayyanchira, line 115: returnToInbox() should happen in all cases right? why not just add it at the bottom common to all cases?

Ayyanchira (commented):
Added some comment

ts/IterableInboxMessageDisplay.tsx
evangreer91: Yes. I can add returnToInbox() to the bottom for all cases.

ts/IterableInboxMessageDisplay.tsx
evangreer91: Yes. I will update.

ts/IterableInboxMessageDisplay.tsx
Ayyanchira: Or just once at the end of the method? 
Is there any case we want to exclude where returning to inbox is not expected?

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -86,20 +96,24 @@ const IterableInboxMessageDisplay = ({
...
       } else if(URL === 'iterable://dismiss') {
          returnToInbox()
+         Iterable.trackInAppClose(rowViewModel.inAppMessage, IterableInAppLocation.inbox, IterableInAppCloseSource.link)
       } else {
          if(Iterable.savedConfig.urlHandler) {
             let action = new IterableAction("openUrl", URL, "")
             let source = IterableActionSource.inApp
             let context = new IterableActionContext(action, source)
 
             Iterable.savedConfig.urlHandler(event.nativeEvent.data, context)
```
roninopf, line 114: Use `URL` instead of `event.nativeEvent.data`?

evangreer91: Yes. I will update.

PR #265, opened 2021-11-10
[MOB-3693] suppress html deep links

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3693](https://iterable.atlassian.net/browse/MOB-3693)

## ✏️ Description

This pull request includes updating the injected javascript to the webview component to disable HTML links for the general case and sending the message to the webview by setting the class of the HTML element. The handleHTMLmessage function is updated to handle external links as well.


ts/IterableInboxMessageDisplay.tsx
```diff
@@ -62,28 +62,15 @@ const IterableInboxMessageDisplay = ({
...
-            } else {
-               window.ReactNativeWebView.postMessage(link.href)
-            }
+            window.ReactNativeWebView.postMessage(link.class)   
          })
-
-         if(link.href === "iterable://dismiss" || link.href === "iterable://delete") {
-            link.href = "javascript:void(0)"
-         }
       })
```
Ayyanchira, line 74: Great refactoring :)

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -107,16 +94,11 @@ const IterableInboxMessageDisplay = ({
...
-      returnToInbox()
-   }
-       
-   const openExternalURL = (event: any) => {
-      if (event.url.slice(0, 4) === 'http') {
-         Linking.openURL(event.url)
-         returnToInbox()
-         return false
+      if (URL.slice(0, 4) === 'http') {
+         Linking.openURL(URL)
```
Ayyanchira, line 98: Are we not using `Linking.canOpenUrl`?

Ayyanchira (commented):
Great refactor of code. Added another comment

ts/IterableInboxMessageDisplay.tsx
evangreer91: Using canOpenUrl() returns a promise that needs to be processed. I thought using this would be more straight forward and accomplish what we want.

evangreer91: Actually we can use canOpenUrl(). I will make the update.

evangreer91: Using canOpenUrl() causes issue with the error links. I think we should go with the current approach.

Ayyanchira (approved):

roninopf (approved):

PR #268, opened 2021-11-12
Update email handlers

griffithcaleb:
## ✏️ Description
The updateEmail method  in the RN SDK does pass down the `onFailureHandler` or `onSuccessHandler` that are available in 
the Native SDKs -> expose these methods to the RN side. 
**Scenario and Use Case**
When launching the app the an 'anonymous' user is created using the `setUserId` method. In the current implementation if a user logs in we will want to `updateEmail` or `setEmail` based on if the user's is currently registered with Iterable. Because there is no way to know this before, we will have to assume that we will want to update the user's email and associate a 'non-placeholder email address with that user'. However if we try and `updateEmail ` on a user's email address that already exists in Iterable the `updateEmail` call will fail silently because the `onFailureHandler` has not been passed down. 

Resolve a promise back across the RN bridge with the message from the failed API call. So that we can then
 'setUserId(null)'  && `setEmail({theEmailTheUserHasLoggedinWith})`


roninopf:
Hi @griffithcaleb, thanks for the PR! Sorry for delays on reviewing and merging, it'll be on our list after the holiday break.

PR #269, opened 2021-11-16
[MOB-3377] consolidate message components into IterableInboxMessageCell component

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3377] (https://iterable.atlassian.net/browse/MOB-3377)

## ✏️ Description

This pull request consolidates the IterableInboxSwipeableRow, IterableInboxClickableRow, and IterableInboxMessageListItem into IterableInboxCell component.

[MOB-3377]: https://iterable.atlassian.net/browse/MOB-3377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ts/IterableInboxMessageCell.tsx
```diff
@@ -139,17 +137,146 @@ const defaultMessageListLayout = (
    )
 }
 
-const IterableInboxMessageListItem = ({
+type MessageCellProps = {
+   key: string,
```
roninopf, line 141: Can you remove unused properties as a result of this consolidation? `key`, at least, isn't used anywhere else.

evangreer91: Yes, thanks.

ts/IterableInboxMessageCell.tsx
```diff
@@ -1,25 +1,23 @@
 'use strict'
 
-import React from 'react'
+import React, { useRef } from 'react'
```
Ayyanchira, line 3: Just a nit pick.
Not sure whats the best practice. Should we write all the modules within same package in curly braces? some outside and some inside looks unformatted.

evangreer91: This is the convention. If I include React within the braces, I get an error that React can only be imported as a default import.

Ayyanchira (approved):
Tested.. Runs on device too..

PR #278, opened 2021-12-28
[MOB-3826] address initial impression tracking

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3826](https://iterable.atlassian.net/browse/MOB-3826)

## ✏️ Description

This pull request includes the functionality to allow the initial message impression to be tracked and sent to the backend. The initial message impressions are calculated based on the message row height and the message list height. The initial message impressions are sent to the backend when the inbox session is started.

### DIFF
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -92,7 +92,7 @@ const IterableInbox = ({
             inboxDataModel.startSession(visibleMessageImpressions)
          } else if(appState === 'background' && Platform.OS === 'android' || appState === 'inactive') {
             inboxDataModel.endSession()
-         } 
+         }
       }
    }, [appState])
 
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -19,7 +19,6 @@ import {
    InboxRowViewModel,
    IterableInboxCustomizations,
    IterableInAppDeleteSource,
-   IterableInAppMessage,
    useAppStateListener,
    useDeviceOrientation,
    Iterable,
@@ -60,7 +59,7 @@ const IterableInbox = ({
    const [animatedValue] = useState<any>(new Animated.Value(0))
    const [isMessageDisplay, setIsMessageDisplay] = useState<boolean>(false)
 
-   const [visibleRowImpressions, setVisibleRowImpressions] = useState<InboxImpressionRowInfo[]>()
+   const [visibleMessageImpressions, setVisibleMessageImpressions] = useState<InboxImpressionRowInfo[]>()
    
    let {
       loadingScreen,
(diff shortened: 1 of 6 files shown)


ts/IterableInbox.tsx
```diff
@@ -109,6 +111,42 @@ const IterableInbox = ({
...
+      } else {
+         inboxDataModel.endSession()
+      }
+   }
+
+   function getInitialMessageImpressions(messageListHeight: number, messageRowHeight: number, rowViewModels: InboxRowViewModel[]) {
+      let numMessages = Math.floor(messageListHeight / messageRowHeight)
+      let visibleRowViewModels = rowViewModels.slice(0, numMessages)
+
+      setVisibleRowImpressions(visibleRowViewModels.map(rowViewModel => {
```
roninopf, line 127: The translation inside that `map` should definitely get separated into its own function.

roninopf: I realized I wrote `var inAppMessage = IterableInAppMessage.fromViewToken(viewToken)` exactly to replace the giant block of text. Is there a reason to not use that?

ts/IterableInbox.tsx
```diff
@@ -109,6 +111,42 @@ const IterableInbox = ({
...
+      if(isFocused) {
+         inboxDataModel.startSession()
+         inboxDataModel.updateVisibleRows(visibleRowImpressions)
+      } else {
+         inboxDataModel.endSession()
+      }
+   }
+
+   function getInitialMessageImpressions(messageListHeight: number, messageRowHeight: number, rowViewModels: InboxRowViewModel[]) {
+      let numMessages = Math.floor(messageListHeight / messageRowHeight)
```
roninopf, line 124: I think this way of calculating the on screen messages is going to be too fragile to use. Is there a way we can skip depending on calculations like this? An easy this would break would be if individual messages have different heights, like different kinds of messages (and subsequently, different heights than the standard) in an inbox.

ts/IterableInbox.tsx
evangreer91: Yes, sounds good.

ts/IterableInbox.tsx
evangreer91: Yes, I did more research and found a way to trigger the onViewableItemsChanged using recordInteraction(). see [here](https://reactnative.dev/docs/flatlist).

ts/IterableInboxMessageList.tsx
```diff
@@ -82,11 +85,11 @@ const IterableInboxMessageList = ({
 
    const inboxSessionItemsChanged = useCallback((
       (info: {viewableItems: Array<ViewToken>, changed: Array<ViewToken>}) => {
-         // logCurrentVisibleRows(info)
+         logCurrentVisibleRows(info)
```
Ayyanchira, line 88: Use Iterablelogger inside logCurrentVisibleRows and use verbose mode so that its not executed when app is published.

Ayyanchira (approved):
Looks fine. Added one comment regarding console logging. logging inside for loop can go heavy on processor. If not in this PR, we should definitely look all console log statements in project together later and make sure its debug or verbose

ts/IterableInboxMessageList.tsx
roninopf: Oh, I wrote that method, I should note it was ONLY meant for debugging, not for production usage. So we should delete it before release.

roninopf: @evangreer91 feel free to just get rid of that function.

evangreer91: I will delete that function and the function call. Thanks.

PR #279, opened 2022-01-04
[MOB-3864] update slider height with customization and custom layout

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-3864] (https://iterable.atlassian.net/browse/MOB-3864)

## ✏️ Description

This pull request includes a small change that allows the delete slider to match the message row height when the styling customized. If the default inbox component is used, the delete slider has a default height. Also, I passed the message row height from the custom layout as a return value that is used by the IterableInboxMessageCell component.

[MOB-3864]: https://iterable.atlassian.net/browse/MOB-3864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ts/IterableInboxMessageCell.tsx
+++ b/ts/IterableInboxMessageCell.tsx
@@ -166,6 +166,12 @@ const IterableInboxMessageCell = ({
 }: MessageCellProps) => {
    const position = useRef(new Animated.ValueXY()).current
 
+   let deleteSliderHeight = customizations.messageRow?.height ? customizations.messageRow.height: 120
+   
+   if(messageListItemLayout(last, rowViewModel)) {
+      deleteSliderHeight = messageListItemLayout(last, rowViewModel)[1]
+   }
+
    const styles = StyleSheet.create({
       textContainer: {
          width: '100%',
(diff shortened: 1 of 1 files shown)


ts/IterableInboxMessageCell.tsx
```diff
@@ -166,6 +166,37 @@ const IterableInboxMessageCell = ({
 }: MessageCellProps) => {
    const position = useRef(new Animated.ValueXY()).current
 
+   let deleteSliderHeight = customizations.messageRow?.height ? customizations.messageRow.height: 120
+   
+   if(messageListItemLayout(last, rowViewModel)) {
+      deleteSliderHeight = messageListItemLayout(last, rowViewModel)[1]
```
Ayyanchira, line 172: is this condition for last row? what is specific about it?

evangreer91: If there is no specified custom row height, it defaults to a height of 120. The issue we were having is when the message row height is customized, the delete slider height was remaining the default of 120.

evangreer91: This is for all message rows. I added the messageRow height as a value passed down to the IterableInboxMessageCell component (messageListItemLayout[0] is the custom layout, messageListItemLayout[1] is the custom message row height). This will be used to update the deleteSlider height in the IterableInboxMessageCell component. The dev app is updated to reflect this. If no custom message row height is provided, the delete slider will default to 120. Let me know if you need to sync for further explanation.

Ayyanchira (approved):
LFTM

PR #284, opened 2022-01-21
[MOB-3923] RN SDK allowed protocols

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-3923](https://iterable.atlassian.net/browse/MOB-3923)

## ✏️ Description

* adds the allowed protocols list to the RN SDK's config, and connects it to the respective native platforms


Ayyanchira (approved):
Verified functionality on Android:
Tested the case:
1. urls not in allowedProtocols are not sent to urlHandler.
2. Added "http"  in allowedProtocols array while setting Iterable config and was able to see urlHandler passed in a value.
LGTM. :)

PR #294, opened 2022-02-16
[MOB-4037] adds delay to deleteRow call to allow message display to be dismissed

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-4037] (https://iterable.atlassian.net/browse/MOB-4037)

## ✏️ Description

This pull request updates the return to inbox functionality, so that for all actions involving returning to the inbox the animation completes first. For example, when a user selects the delete button, the message display closes first before the message is deleted.

[MOB-4037]: https://iterable.atlassian.net/browse/MOB-4037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ts/IterableInbox.tsx
+++ b/ts/IterableInbox.tsx
@@ -8,7 +8,8 @@ import {
    Animated,
    NativeModules,
    NativeEventEmitter,
-   Platform
+   Platform,
+   Linking
 } from 'react-native'
 
 import {
@@ -18,6 +19,7 @@ import {
    IterableInAppDeleteSource,
    IterableInAppLocation,
    IterableInboxEmptyState,
+   IterableActionContext,
    InboxImpressionRowInfo
 } from '.'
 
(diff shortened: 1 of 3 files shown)


ts/IterableInbox.tsx
```diff
@@ -165,8 +165,13 @@ const IterableInbox = ({
       fetchInboxMessages()
    }
 
-   function returnToInbox() {
-      reset()
+   function returnToInbox(callback: Function) {
```
roninopf, line 168: Is there a way to let this parameter be optional, so that if nothing is entered, the function doesn't do any callback? It would make [this call](https://github.com/Iterable/react-native-sdk/pull/294/files#diff-f4c0dd9039a860fa23d99f3faa1f7bbd72be7240d8dcc0117824979ec837b051R117) more intuitive if it was just `returnToInbox()`, instead of needing a blank callback.

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -92,36 +92,29 @@ const IterableInboxMessageDisplay = ({
...
+      } else if(URL === 'iterable://dismiss') {
+         returnToInbox(() => {})
+      } else if (URL.slice(0, 4) === 'http') {
+         returnToInbox(() => Linking.openURL(URL))
+      } else {
+         returnToInbox(() => {
+            if(Iterable.savedConfig.urlHandler) {
+               Iterable.savedConfig.urlHandler(URL, context)
+            } 
+         })
```
roninopf, line 108: I'm thinking this function (`handleHTMLMessage`) may need to be named differently now. It doesn't really connect to what it does, which is handling all actions from an in-app.

Ayyanchira (approved):
Good 👍

Issue #307, opened 2022-03-15
iOS Deployment Target needs updating


hytea:
I am getting an error when trying to build my app for iOS on the latest react native SDK version the React Native SDK has an iOS deployment target of 9.0 but the Iterable iOS SDK has an iOS deployment target of 10.0. 

If I update the deployment target manually it works but it gets reset anytime I deepclean the project.

ivansnek:
> I am getting an error when trying to build my app for iOS on the latest react native SDK version the React Native SDK has an iOS deployment target of 9.0 but the Iterable iOS SDK has an iOS deployment target of 10.0.
> 
> If I update the deployment target manually it works but it gets reset anytime I deepclean the project.

Having the same issue here! 🤚🏽 

@hytea how did you manually change this? does it work for release?

roninopf:
Hi! Sorry we missed this, it's since been fixed in version 1.2.1. If there's a new issue with that, please feel free to open a new ticket!

PR #308, opened 2022-03-22
Update iOS Deployment Target

jgrana:
This fixes issue #307  

## 🔹 JIRA Ticket(s) if any

No JIRA ticket

## ✏️ Description

Updates the deployment target to fix builds. Spec still needs to be pushed to cocoapods


roninopf:
Thanks for the submission @jgrana ! I'm testing this out, but this looks like the right fix, I'll proceed with merging and then releasing a new version when it's in.

roninopf (approved):

Issue #356, opened 2022-05-12
Issue after upgrading to 1.3.1


diego-paired:
Immediately after upgrading from 1.3.0 to 1.3.1, the following error prevents the app from booting:
```
TypeError: Cannot read property 'prototype' of undefined
    at ?anon_0_ (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=io.betterhalf.app:151379:101)
```

```
System:
    OS: macOS 12.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 108.48 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
    Watchman: 2021.09.27.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: Not Found
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.4, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK:
      API Levels: 25, 27, 28, 29, 30, 31, 32
      Build Tools: 30.0.2, 30.0.3, 31.0.0, 32.0.0, 32.0.0, 33.0.0
      System Images: android-31 | Google APIs ARM 64 v8a, android-32 | Google APIs ARM 64 v8a, android-Sv2 | Google APIs ARM 64 v8a
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8309675
    Xcode: 13.3.1/13E500a - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.13 - /Users/diegolopez/.sdkman/candidates/java/current/bin/javac
  npmPackages:
    @react-native-community/cli: ^7.0.1 => 7.0.3
    react: ^17.0.2 => 17.0.2
    react-native: ^0.68.2 => 0.68.2
```

alex-a-pereira:
[...]

PR #357, opened 2022-05-13
ready: Add missing Iterable and IterableConfig exports back to index.ts

alex-a-pereira:
## 🔹 JIRA Ticket(s) if any

No tickets - PR is to address this issue https://github.com/Iterable/react-native-sdk/issues/356

## ✏️ Description

After the 1.3.1 release, `Iterable` and `IterableConfig` were no longer exported from index.ts. This PR adds them back in. I've also made the unit tests import from index.ts rather than the implementation files, as consumers will be importing from index (so if this happens again, it will be caught by unit tests in CI).

---

If these removals were intentional and there's a new way to import these classes from the SDK, feel free to close out this PR. Documentation would need to be updated in that case as found here: https://support.iterable.com/hc/en-us/articles/360045714132-Installing-Iterable-s-React-Native-SDK-#_5-import-classes-as-needed

### DIFF
--- a/ts/__tests__/Iterable.spec.ts
+++ b/ts/__tests__/Iterable.spec.ts
@@ -3,8 +3,13 @@ import { MockLinking } from '../__mocks__/MockLinking'
 import { TestHelper } from './TestHelper'
 import { NativeEventEmitter } from 'react-native'
 
+// import from the same location that consumers import from
 import {
   Iterable,
+  IterableConfig
+} from '../index'
+
+import {
   IterableAttributionInfo,
   IterableCommerceItem,
   IterableActionContext,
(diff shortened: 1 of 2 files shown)


evangreer91 (approved):

alex-a-pereira:
@evangreer91 thanks, couple of questions
1. do you want me to bump the version to `1.3.2` in this PR ???
2. re: failing integration tests, anything I can do here to get tests to run properly before releasing to NPM?

I think integration tests will still fail after this PR is merged. `integration-testing/ts/App.tsx` also imports `IterableInAppMessage` from the package which is not exported. I can update my PR and also export that

roninopf:
@alex-a-pereira Thanks for catching this, don't worry about the release or bumping the version, since there's always a possibility of other things taking precedence (unlikely for this, but generally speaking), or being bundled in with other fixes.

We've got some issues with integration tests in general, which I'm handling separately.

clemencov:
It seems that `1.3.2` with the fix was never published.

PR #360, opened 2022-05-18
[MOB-2572] create a logger class and have option to disable logging

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-2572](https://iterable.atlassian.net/browse/MOB-2572)

## ✏️ Description

* creates `IterableLogger` specific to the RN SDK
* adds a boolean option in `IterableConfig` (default: `true`) to always print
* replaces anything in our SDK that previously used `console.log` to go through `IterableLogger`

### DIFF
--- a/ts/Iterable.ts
+++ b/ts/Iterable.ts
@@ -394,7 +394,7 @@ class Iterable {
     subscribedMessageTypeIds: Array<number> | undefined,
     campaignId: number,
     templateId: number) {
-    console.log("updateSubscriptions")
+    Iterable.logger.log("updateSubscriptions")
     RNIterableAPI.updateSubscriptions(emailListIds, unsubscribedChannelIds, unsubscribedMessageTypeIds, subscribedMessageTypeIds, campaignId, templateId)
   }
 
--- a/ts/Iterable.ts
+++ b/ts/Iterable.ts
@@ -134,9 +134,9 @@ class Iterable {
   * @param {IterableConfig} config
   */
   static initialize(apiKey: string, config: IterableConfig = new IterableConfig()): Promise<boolean> {
-    Iterable.logger.log("initialize: " + apiKey)
-
     Iterable.savedConfig = config
+
+    Iterable.logger.log("initialize: " + apiKey)
 
     this.setupEventHandlers()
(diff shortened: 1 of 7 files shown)


ts/IterableLogger.ts
```diff
@@ -0,0 +1,15 @@
+import { Iterable } from "./Iterable"
+
+class IterableLogger {
+    log(message: String) {
+        // default to `true` in the case of unit testing where `Iterable` is not initialized
+        // which is most likely in a debug environment anyways
+        var loggingEnabled = Iterable.savedConfig?.logReactNativeSdkCalls ?? true
+
+        if (loggingEnabled) {
```
Ayyanchira, line 9: Should we not have different levels on RN side as well? Like INFO, Debug , Error?

roninopf: We can support that in a future version, but for now, we don't exactly have that distinction either, since the RN SDK doesn't pass along errors from the native layer, either.

ts/Iterable.ts
```diff
@@ -353,7 +374,8 @@ class Iterable {
   * @param {string} link URL in string form to be either opened as an app link or as a normal one
   */
   static handleAppLink(link: string): Promise<boolean> {
-    console.log("handleAppLink")
+    Iterable.logger.log("handleAppLink")
+
     return RNIterableAPI.handleAppLink(link)
   }
```
evangreer91, line 381: There is a console.log on line 397 that should be converted.

roninopf: Nice catch! It didn't select the correct one but I found it.

evangreer91 (approved):

hetmann:
@roninopf we're getting app crashes due to `TypeError: Cannot read property 'log' of undefined` when using `getEmail` or `setEmail` features.

Can you double check the current implementation?

Kyonru:
> @roninopf we're getting app crashes due to `TypeError: Cannot read property 'log' of undefined` when using `getEmail` or `setEmail` features.
> 
> Can you double check the current implementation?

It's probably an issue when trying any of the static methods like `getEmail` or `setEmail` before initializing the sdk. (I had the same issue).

PR #375, opened 2022-07-12
fix gradle config to inherit values from parent project using safextget

shamilovtim:
## 🔹 JIRA Ticket(s) if any

N/A external

## ✏️ Description

Currently we are having to patch the plugin because the gradle build config is hardcoding values rather than inheriting them from the parent project. I applied the standard `safeExtGet` code popular in most gradle configs to inherit the parent values with a fallback. Otherwise user's projects will error out due to the hardcoding and users must patch the plugin to keep building.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,12 +2,16 @@ apply plugin: 'com.android.library'
 
 import groovy.json.JsonSlurper
 
+def safeExtGet(prop, fallback) {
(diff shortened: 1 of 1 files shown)


shamilovtim:
What's it take to merge this? @roninopf

shamilovtim:
@evangreer91 could you please merge this? It's harmless and would eliminate a patch file for us.

PR #378, opened 2022-07-18
[MOB-4449] update email with token

roninopf:
## 🔹 JIRA Ticket(s) if any

* [MOB-4449](https://iterable.atlassian.net/browse/MOB-4449)

## ✏️ Description

* adds RN bindings for `updateEmail` with token

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -113,9 +113,9 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     }
 
     @ReactMethod
-    public void updateEmail(String email) {
+    public void updateEmail(String email, @Nullable String authToken) {
         IterableLogger.d(TAG, "updateEmail: " + email);
-        IterableApi.getInstance().updateEmail(email);
+        IterableApi.getInstance().updateEmail(email, authToken);
     }
 
     @ReactMethod
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -113,6 +113,13 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     }
 
     @ReactMethod
+    public void setEmail(@Nullable String email, @Nullable String authToken) {
+        IterableLogger.d(TAG, "setEmail: " + email + " authToken: " + authToken);
+
+        IterableApi.getInstance().setEmail(email, authToken);
+    }
+
+    @ReactMethod
     public void updateEmail(String email, @Nullable String authToken) {
         IterableLogger.d(TAG, "updateEmail: " + email);
         IterableApi.getInstance().updateEmail(email, authToken);
@@ -133,6 +140,13 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     public void setUserId(@Nullable String userId) {
         IterableLogger.d(TAG, "setUserId: " + userId);
         IterableApi.getInstance().setUserId(userId);
+    }
+
+    @ReactMethod
+    public void setUserId(@Nullable String userId, @Nullable String authToken) {
+        IterableLogger.d(TAG, "setUserId: " + userId + " authToken: " + authToken);
+        
+        IterableApi.getInstance().setUserId(userId, authToken);
     }
 
     @ReactMethod
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -120,8 +120,16 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
     }
 
     @ReactMethod
-    public void updateEmail(String email, @Nullable String authToken) {
+    public void updateEmail(String email) {
         IterableLogger.d(TAG, "updateEmail: " + email);
+
+        IterableApi.getInstance().updateEmail(email);
+    }
+
+    @ReactMethod
+    public void updateEmail(String email, @Nullable String authToken) {
+        IterableLogger.d(TAG, "updateEmail: " + email + " authToken: " + authToken);
+
         IterableApi.getInstance().updateEmail(email, authToken);
     }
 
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -113,9 +113,9 @@ public void setEmail(@Nullable String email) {
     }
 
     @ReactMethod
-    public void updateEmail(String email) {
+    public void updateEmail(String email, @Nullable String authToken) {
```
Ayyanchira, line 116: There should be new method for the one's with authToken right? Old method should still be there. Otherwise it will be a breaking change

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
```diff
@@ -113,9 +113,9 @@ public void setEmail(@Nullable String email) {
...
         IterableLogger.d(TAG, "updateEmail: " + email);
-        IterableApi.getInstance().updateEmail(email);
+        IterableApi.getInstance().updateEmail(email, authToken);
```
Ayyanchira, line 118: Might be mistaken but I am not seeing updateEmail method with authToken in our Android SDK. Its just setEmail and setUserID.. strange coz i felt I've seen it recently 😅

ts/Iterable.ts
```diff
@@ -362,11 +362,12 @@ class Iterable {
...
   * 
   * @param email the new email to set
+  * @param authToken the new auth token (JWT) to set with the new email, optional - if null/undefined, no JWT related action will be taken
   */
-  static updateEmail(email: string) {
+  static updateEmail(email: string, authToken?: string | undefined) {
     Iterable.logger.log("updateEmail")
 
-    RNIterableAPI.updateEmail(email)
+    RNIterableAPI.updateEmail(email, authToken)
```
Ayyanchira, line 370: Same like above.. instead of replacing, it can be just addition of new method?

roninopf: It's defined with `string | undefined` so both ways work!

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
roninopf: Yeah, I just put it in ahead of time.

android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
roninopf: Added! Forgot I had to do that.

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -301,11 +315,15 @@ class ReactIterableAPI: RCTEventEmitter {
         IterableAPI.updateUser(dataFields, mergeNestedObjects: mergeNestedObjects)
     }
     
-    @objc(updateEmail:)
-    func updateEmail(email: String) {
+    @objc(updateEmail:authToken:)
+    func updateEmail(email: String, with authToken: String?) {
```
evangreer91, line 319: Is this function call supposed to have a with keyword? Seems unusual

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -97,6 +97,13 @@ class ReactIterableAPI: RCTEventEmitter {
         
         IterableAPI.email = email
     }
+
+    @objc(setEmail:withAuthToken:)
+    func set(email: String?, authToken: String?) {
+        ITBInfo()
```
Ayyanchira, line 103: Its good swift does not log the email and authToken. In future, Android can have this approach as well 👍🏼

ts/Iterable.ts
```diff
@@ -167,10 +167,10 @@ class Iterable {
   * Set the user of the SDK by email address
   * @param {string | undefined} email the email address of the user
   */
-  static setEmail(email: string | undefined) {
+  static setEmail(email: string | undefined, authToken?: string | undefined) {
```
Ayyanchira, line 170: I'm assuming, extra argument in RN is optional and can be passed just one parameter as well. This is cool that theres no need of extra methods.

Ayyanchira (approved):
LGTM

ios/RNIterableAPI/RNIterableAPI.m
```diff
@@ -80,7 +80,8 @@ @interface RCT_EXTERN_REMAP_MODULE(RNIterableAPI, ReactIterableAPI, NSObject)
 RCT_EXTERN_METHOD(updateUser: (nonnull NSDictionary *) dataFields
                   mergeNestedObjects: (BOOL) mergeNestedObjects)
 
-RCT_EXTERN_METHOD(updateEmail: (nonnull NSString *) email)
+RCT_EXTERN_METHOD(updateEmail: (nonnull NSString *) email
+                  authToken: (NSString *) authToken)
```
evangreer91, line 84: Is there a reason setEmail and setUserId are not updated here?

ts/__tests__/Iterable.spec.ts
```diff
@@ -136,12 +136,21 @@ test("update user", () => {
...
+})
+
+test("update email (with token)", () => {
+  const newEmail = "woo@newemail.com"
+  const newToken = "token2"
+
+  Iterable.updateEmail(newEmail, newToken)
+
+  expect(MockRNIterableAPI.updateEmail).toBeCalledWith(newEmail, newToken)
 })
```
evangreer91, line 154: So there are no tests for setEmail and setUserId with token?

ios/RNIterableAPI/ReactIterableAPI.swift
Ayyanchira: haha yes..  thats how iOS functions are written to make function names more understandable...

ts/Iterable.ts
roninopf: Yep, you can see our other API functions, same thing going on there.

ts/__tests__/Iterable.spec.ts
roninopf: Yeah, not yet. I tried to set it up, but it was more complicated than expected.

ios/RNIterableAPI/ReactIterableAPI.swift
roninopf: Ah yeah, all we log is the function call

ios/RNIterableAPI/RNIterableAPI.m
roninopf: Forgot it! Good catch.

ts/__tests__/Iterable.spec.ts
evangreer91: got it

evangreer91 (approved):

PR #387, opened 2022-08-12
[MOB-4703] Fix action source deserialization

vbabenkoru:
## 🔹 JIRA Ticket(s) if any

* [MOB-4703](https://iterable.atlassian.net/browse/MOB-4703)

## ✏️ Description

We're serializing the context with `source` as the field name, but were using `actionSource` when deserializing, so the field was always `undefined`.


Ayyanchira (approved):
Good fix :) 👍🏼

PR #398, opened 2022-10-06
Mob 4695 authhandler support for success and failure callbacks

Ayyanchira:
## 🔹 JIRA Ticket(s) if any

* [MOB-4695](https://iterable.atlassian.net/browse/MOB-4695)

## ✏️ Description

1. case handleAuthSuccessCalled and handleAuthFailureCalled added to both Android and iOS Native bridge.
2. If promise type is of type AuthResponse, we will look for string inside it and callbacks if present will be called after waiting for 1 second on Android. For iOS, semaphores is already implemented.
3. If promise if of type ""(string), code should work like before and no callbacks come into the scene.

TODO: Have to remove console log statements which are unnecessary

Verified the functionality of callbacks being called on Android and iOS.

### DIFF
--- a/ios/RNIterableAPI/ReactIterableAPI.swift
+++ b/ios/RNIterableAPI/ReactIterableAPI.swift
@@ -33,6 +33,8 @@ class ReactIterableAPI: RCTEventEmitter {
         case handleInAppCalled
         case handleAuthCalled
         case receivedIterableInboxChanged
+        case handleAuthSuccessCalled
+        case handleAuthFailureCalled
     }
     
     override func supportedEvents() -> [String]! {
@@ -662,12 +664,18 @@ extension ReactIterableAPI: IterableAuthDelegate {
                 DispatchQueue.main.async {
                     completion(self.passedAuthToken)
                 }
+                
+                self.sendEvent(withName: EventName.handleAuthSuccessCalled.rawValue,
+                               body: nil)
             } else {
                 ITBInfo("authTokenRetrieval timed out")
                 
                 DispatchQueue.main.async {
                     completion(nil)
                 }
+                
+                self.sendEvent(withName: EventName.handleAuthFailureCalled.rawValue,
+                               body: nil)
             }
         }
     }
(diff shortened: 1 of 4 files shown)


mprew97 (commented):
Left a few comments but awesome job getting to a solution here 💯

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -662,12 +664,18 @@ extension ReactIterableAPI: IterableAuthDelegate {
                 DispatchQueue.main.async {
                     completion(self.passedAuthToken)
                 }
+                
+                self.sendEvent(withName: EventName.handleAuthSuccessCalled.rawValue,
```
davidtruong, line 668: Do we need .rawValue here or would it automatically map to correctly without it?

ts/Iterable.ts
```diff
@@ -451,13 +458,52 @@ class Iterable {
...
+                Iterable.logger.log('Unexpected promise returned. Auth token expects promise of String or AuthResponse type.')
+              }
+            }).catch(e => Iterable.logger.log(e))
+        }
+      )
+
+      RNEventEmitter.addListener(
+        EventName.handleAuthSuccessCalled,
+        () => {
+          authResponseCallback = "success"
```
davidtruong, line 500: Can you use the enum you created here to keep it consistent?

ts/Iterable.ts
```diff
@@ -30,6 +30,11 @@ enum IterableActionSource {
   inApp = 2
 }
 
+enum AuthResponseCallback {
+  SUCCESS = 'success',
```
davidtruong, line 34: Do you need a string value for the enum defined?

ios/RNIterableAPI/ReactIterableAPI.swift
Ayyanchira: Just keeping it consistent with other Send events as of now. 
Send events usually take strings and .rawValue can extract that from enum.

ts/Iterable.ts
Ayyanchira: Makes sense. Removed the string value and tested and it still works. Thanks :)

davidtruong (approved):
LGTM

PR #403, opened 2022-10-19
[MOB-4982] adds IterableInAppMessage to index.ts

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-4982](https://iterable.atlassian.net/browse/MOB-4982)

## ✏️ Description

This pull request adds the IterableInAppMessage to the top-level index.ts. Adjusts imports to prevent require cycles.

### DIFF
--- a/ts/IterableInboxMessageDisplay.tsx
+++ b/ts/IterableInboxMessageDisplay.tsx
@@ -18,10 +18,15 @@ import {
    IterableHtmlInAppContent, 
    IterableEdgeInsets,
    IterableInAppLocation,
-   IterableInAppCloseSource,
+   IterableInAppCloseSource
+} from './IterableInAppClasses'
+
+import { 
    IterableAction, 
    IterableActionContext
-} from '.'
+ } from './IterableAction'
+
+import InboxRowViewModel from './InboxRowViewModel'
 
 import { IterableActionSource } from './Iterable'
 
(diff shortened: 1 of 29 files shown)


ts/index.ts
```diff
@@ -6,6 +6,7 @@
 */
 
 import { Iterable } from './Iterable'
+import IterableInAppManager from './IterableInAppManager'
```
Ayyanchira, line 9: nit pick: inconsistent with above line. {} present in line 8 while its not used in line 9

ts/IterableInboxMessageDisplay.tsx
```diff
@@ -6,14 +6,14 @@ import {
    View,
    ScrollView,
    StyleSheet,
-   Platform,
```
Ayyanchira, line 9: Why is platform removed?

Ayyanchira (commented):
Added some comments. 
Also, this might be a breaking change for devs importing new SDK right? Might have to document that as well?

ts/index.ts
evangreer91: IterableInAppManager is a default export, while Iterable is a named export. see [here](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export).

ts/IterableInboxMessageDisplay.tsx
evangreer91: Platform is not used in the component.

Ayyanchira (approved):

PR #406, opened 2022-10-21
[MOB-5028] removes promise processing

evangreer91:
## 🔹 JIRA Ticket(s) if any

* [MOB-5028](https://iterable.atlassian.net/browse/MOB-5028)

## ✏️ Description

This pull request includes removing the promise processing from the SDK. The promise is returned by the SDK and processed on the application side.

### DIFF
--- a/ts/IterableInAppManager.ts
+++ b/ts/IterableInAppManager.ts
@@ -31,7 +31,7 @@ class IterableInAppManager {
   getMessages(): Promise<Array<IterableInAppMessage>> {
     Iterable.logger.log("InAppManager.getMessages")
 
-    return RNIterableAPI.getInAppMessages().then((messages: Array<any>) => messages.map(message => { return IterableInAppMessage.fromDict(message) }))
+    return RNIterableAPI.getInAppMessages()
   }
 
    /**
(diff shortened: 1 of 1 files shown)


pauljung14 (approved):
Lgtm, but is there a PR to update documentation? Sounds like customers would need to be aware of this in next release

ts/IterableInAppManager.ts
```diff
@@ -102,9 +102,6 @@ class IterableInAppManager {
      Iterable.logger.log("InAppManager.getHtmlContentForMessage")
 
     return RNIterableAPI.getHtmlInAppContentForMessage(message.messageId)
-      .then((content: any) => {
-        return IterableHtmlInAppContent.fromDict(content)
-      })
```
Ayyanchira, line 107: Wow..  the `.then` part was unnecessary the whole time? 
Have you verified if its properly returning just the HTML part?

Ayyanchira (approved):
👌🏼 Wow.. thats such a great find Evan!

ts/IterableInAppManager.ts
evangreer91: Yes, but I will verify before merging.

evangreer91:
> Lgtm, but is there a PR to update documentation? Sounds like customers would need to be aware of this in next release

This does not change what is returned by the APIs. So, implementation on the application side is unchanged.

PR #484, opened 2023-04-04
[MOB-5870] add data center configuration

evantk91:
## 🔹 JIRA Ticket(s) if any

* [MOB-5870](https://iterable.atlassian.net/browse/MOB-5870)

## ✏️ Description

- creates the IterableDataRegion enum which will inform the native layer
- adds the dataRegion value to the IterableConfig class
- updates the unit tests for the default config values for the added dataRegion value

[MOB-5870]: https://iterable.atlassian.net/browse/MOB-5870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/com/iterable/reactnative/Serialization.java
+++ b/android/src/main/java/com/iterable/reactnative/Serialization.java
@@ -15,6 +15,7 @@ import com.iterable.iterableapi.CommerceItem;
 import com.iterable.iterableapi.IterableAction;
 import com.iterable.iterableapi.IterableActionContext;
 import com.iterable.iterableapi.IterableConfig;
+import com.iterable.iterableapi.IterableDataRegion;
 import com.iterable.iterableapi.IterableInAppCloseAction;
 import com.iterable.iterableapi.IterableInAppDeleteActionType;
 import com.iterable.iterableapi.IterableInAppHandler;
@@ -196,6 +197,20 @@ class Serialization {
                 }
 
                 configBuilder.setLogLevel(logLevel);
+            }
+
+            if(iterableContextJSON.has("dataRegion")) {
+                int dataRegion = iterableContextJSON.getInt("dataRegion");
+                switch (dataRegion) {
+                    case 1:
+                        dataRegion = IterableDataRegion.EU;
+                        break;
+                    default:
+                        dataRegion = IterableDataRegion.US;
+                        break;
+                }
+
+                configBuilder.setDataRegion(dataRegion);
             }
 
             return configBuilder;
--- a/android/src/main/java/com/iterable/reactnative/Serialization.java
+++ b/android/src/main/java/com/iterable/reactnative/Serialization.java
@@ -201,16 +201,20 @@ class Serialization {
 
             if(iterableContextJSON.has("dataRegion")) {
                 int dataRegion = iterableContextJSON.getInt("dataRegion");
+                IterableDataRegion iterableDataRegion = IterableDataRegion.US;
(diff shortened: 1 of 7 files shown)


Ayyanchira (commented):
A serialization part is missing where this config from Typescript will be converted to respective native configs

android/src/main/java/com/iterable/reactnative/Serialization.java
```diff
@@ -198,6 +199,20 @@ static IterableConfig.Builder getConfigFromReadableMap(ReadableMap iterableConte
                 configBuilder.setLogLevel(logLevel);
             }
 
+            if(iterableContextJSON.has("dataRegion")) {
+                int dataRegion = iterableContextJSON.getInt("dataRegion");
```
Ayyanchira, line 203: Although enums are represented as integers internally, in java its treated as class. So a direct mapping like this is not possible. [More reading](https://stackoverflow.com/a/5878986)

android/src/main/java/com/iterable/reactnative/Serialization.java
```diff
@@ -198,6 +199,20 @@ static IterableConfig.Builder getConfigFromReadableMap(ReadableMap iterableConte
...
+                switch (dataRegion) {
+                    case 1:
+                        dataRegion = IterableDataRegion.EU;
```
Ayyanchira, line 206: 0 is US and 1 is EU in native. Will be good to have an explicit case check for 0. Default can still stay.

android/src/main/java/com/iterable/reactnative/Serialization.java
```diff
@@ -198,6 +199,20 @@ static IterableConfig.Builder getConfigFromReadableMap(ReadableMap iterableConte
...
+                        break;
+                    default:
+                        dataRegion = IterableDataRegion.US;
+                        break;
+                }
+
+                configBuilder.setDataRegion(dataRegion);
```
Ayyanchira, line 213:
```suggestion
 int dataRegionInt = iterableContextJSON.getInt("dataRegionInt");
                IterableDataRegion region;
                switch (dataRegionInt) {
                    case 1:
                        region = IterableDataRegion.EU;
                        break;
                    case 2:
                        region = IterableDataRegion.US;
                    default:
                        region = IterableDataRegion.US;
                        break;
                }

                configBuilder.setDataRegion(region);
```

Ayyanchira: Something like this..

Ayyanchira (commented):
Added some comments

android/src/main/java/com/iterable/reactnative/Serialization.java
evantk91: Makes sense

android/src/main/java/com/iterable/reactnative/Serialization.java
evantk91: Isnt the data region in the RN SDK represented as an integer though?

evantk91: Oh I see... In the Android SDK, we need a getEnum function to get the enum value from the string. Thus, in RN SDK, we should set the value of the data region to the endpoint.

Ayyanchira (approved):
LGTM

PR #488, opened 2023-04-26
[MOB-6055] callback on setEmail/setUserId

devcsomnicg:
## 🔹 JIRA Ticket(s) if any

* [MOB-6055](https://iterable.atlassian.net/browse/MOB-6055)

## ✏️ Description

> Add callbacks for setEmail/setUserId, and it returns success/failure based on device token is registered successfully or not. It will always return success if autoPushRegistration is false.

[MOB-6055]: https://iterable.atlassian.net/browse/MOB-6055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/SampleApp/javascript/js/SettingsTab.js
+++ b/SampleApp/javascript/js/SettingsTab.js
@@ -8,7 +8,9 @@ class SettingsTab extends Component {
         super(props);
         this.onLoginTapped = () => {
             console.log("onLoginTapped");
-            Iterable.setEmail(this.state.email);
+            Iterable.setEmail(this.state.email, (result) => {
+                console.log('setEmail result:'+result)
+            });
             this.updateState();
         };
         this.onLogoutTapped = () => {
--- a/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
+++ b/android/src/main/java/com/iterable/reactnative/RNIterableAPIModule.java
@@ -122,7 +122,7 @@ public class RNIterableAPIModule extends ReactContextBaseJavaModule implements I
         IterableApi.getInstance().setEmail(email, authToken, new SuccessHandler() {
             @Override
             public void onSuccess(@NonNull JSONObject data) {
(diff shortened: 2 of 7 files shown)


harrymash2006:
@roninopf Could you please start review for this ?

devcsomnicg:
@Ayyanchira Could you please start review for this as well ?

SampleApp/javascript/js/SettingsTab.js
```diff
@@ -8,7 +8,9 @@ class SettingsTab extends Component {
         super(props);
         this.onLoginTapped = () => {
             console.log("onLoginTapped");
-            Iterable.setEmail(this.state.email);
+            Iterable.setEmail(this.state.email, (result) => {
+                console.log('setEmail result:'+result)
```
Ayyanchira, line 12: So this result will have both success and failure callback data. And devs will have to basically separate it out here. Right? For typescript, perhaps this could be way to go.. 
@pauljung14 , can you advice if for front end, having one generic callback for success and failure is recommended. 

I'm thinking there is promise - resolve and reject concept. Can it apply here?

pauljung14: Is the only purpose for adding this callback to log whether registering the device token failed?

devcsomnicg: yes actually, we only need to know if it succeeded or not.

devcsomnicg: @pauljung14 any update on this ?

pauljung14 (approved):

SampleApp/javascript/js/SettingsTab.js
pauljung14: sounds fine to me

PR #551, opened 2024-05-10
Guard against emitting events if there are no listeners

andrew-schenk:
## 🔹 JIRA Ticket(s) if any

N/A

## ✏️ Description

When hot refreshing a react-native app, it is possible to cause a crash when the Iterable registration has not completed. This PR adds a simple check to prevent the native module from emitting events if there are no JS listeners.


jferrao-itrbl:
New-arch rewrite already uses `shouldEmit` / `startObserving` in `ReactIterableAPI.swift`.

PR #593, opened 2024-10-23
[MOB-10036] Example app: delete old sample 

lposen:
## 🔹 JIRA Ticket(s) if any

* [MOB-10036](https://iterable.atlassian.net/browse/MOB-10036)

## ✏️ Description

Delete old sample app

[MOB-10036]: https://iterable.atlassian.net/browse/MOB-10036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -47,14 +47,13 @@ To learn more about the SDK, read:
 
 ## Sample projects
 
-For sample code, take a look at the following projects:
+For sample code, take a look at the following project:
 
-- [SampleApp (JavaScript)](https://github.com/Iterable/react-native-sdk/tree/master/SampleApp/javascript)
-- [SampleApp (TypeScript)](https://github.com/Iterable/react-native-sdk/tree/master/SampleApp/typescript)
+- [Example App](https://github.com/Iterable/react-native-sdk/tree/master/example)
```
evantk91, line 52: For the example app, is there a reason we only have one version of example app? I think there was a reason we had a typescript version and javascript version. We might need to think through providing one of each again.

README.md
```diff
@@ -47,14 +47,13 @@ To learn more about the SDK, read:
...
 
 ## Version mapping
 
-For quick reference, the following table lists the versions of the [Android SDK](https://github.com/Iterable/iterable-android-sdk) and the [iOS SDK](https://github.com/Iterable/swift-sdk) the React Native SDK points to for version 1.2.0 and above. 
+For quick reference, the following table lists the versions of the [Android SDK](https://github.com/Iterable/iterable-android-sdk) and the [iOS SDK](https://github.com/Iterable/swift-sdk) the React Native SDK points to for version 1.2.0 and above.
```
evantk91, line 56: This should be updated since the iOS SDK has been renamed. (iOS SDK)[https://github.com/Iterable/iterable-swift-sdk].

package.json
```diff
@@ -156,8 +156,7 @@
   "eslintIgnore": [
     "node_modules/",
     "lib/",
-    "integration-testing/",
-    "SampleApp/"
+    "integration-testing/"
```
evantk91, line 159: Do we want example app to be ignored by es lint similar to sample app?

tsconfig.json
```diff
@@ -26,5 +26,5 @@
     "target": "ESNext",
     "verbatimModuleSyntax": true
   },
-  "exclude": ["SampleApp/", "integration-testing/"]
+  "exclude": ["integration-testing/"]
```
evantk91, line 29: Here too. I feel like we just want to replace SampleApp with example.

evantk91 (commented):
Left a few clarifications.

README.md
lposen: I spoke briefly to @Ayyanchira about this.  This is not a pattern I've ever seen in frontend development -- I'd venture to say that it actually makes it more difficult to understand.  I definitely would not use this pattern.

package.json
lposen: Nope!  I just ignored it previously as I didn't want to go through and fix all the issues in something that I knew was going to be deleted.

tsconfig.json
lposen: Same as the above comment -- it shouldn't be added here as we do want it to be type checked.  `integration-testing` is in here as it is going to be deleted later, so I'm not going to spend time typing it.

README.md
lposen: Updated

README.md
evantk91: Makes sense.

evantk91 (approved):

PR #596, opened 2024-11-01
ready: eslint deps are devDependencies

alex-a-pereira:
## Issue

Addresses this issue that I opened earlier today

https://github.com/Iterable/react-native-sdk/issues/595

## ✏️ Description

Move `eslint` and `eslint-config-airbnb` dependencies into the `devDependencies` object in `package.json`.

Thie prevents `yarn` from printing warnings:
```
➤ YN0002: │ @iterable/react-native-sdk@npm:1.3.20 [b01c5] doesn't provide eslint-plugin-import (p67cd9), requested by eslint-config-airbnb
➤ YN0002: │ @iterable/react-native-sdk@npm:1.3.20 [b01c5] doesn't provide eslint-plugin-jsx-a11y (p16ac4), requested by eslint-config-airbnb
➤ YN0002: │ @iterable/react-native-sdk@npm:1.3.20 [b01c5] doesn't provide eslint-plugin-react (pf10d0), requested by eslint-config-airbnb
➤ YN0002: │ @iterable/react-native-sdk@npm:1.3.20 [b01c5] doesn't provide eslint-plugin-react-hooks (p53156), requested by eslint-config-airbnb
```

PR #603, opened 2024-11-15
Move dependencies to peer dependencies

johnhaup:
## ✏️ Description

> Using the UI components (example: Inbox) breaks because react navigation and safe area context are dependencies in the SDK, so they get installed in the package's own node_modules (`node_modules/@iterable/react-native-sdk/node_modules`).  The makes the components look for navigation context and safe area context from an instance in its own node_modules, instead of the consumer's project.

- Removed react-hooks package (unused in SDK)
- Moved lint packages from `dependencies` to `devDependencies`
- Moved packages from `dependencies` to `peerDependencies`, using major versions as minimum


lposen (approved):

lposen:
Approved.  Looks good to me.

johnhaup:
This also address #596

bhandanyan-nomad:
@lposen @johnhaup exciting change. I'd love to incorporate this into my project because I'm looking to upgrade `react-native-safe-area-context` from v4 to v5 and this change will allow me to avoid having two versions installed or needing to add a manual resolution to the version of my choosing.

Can you let me know if this will be available in a new release soon (ie this week) or not?

alex-a-pereira:
thank you!!

PR #771, opened 2025-10-31
feat: make @react-navigation/native dependency optional

### LINKED ISSUE #770: Implicit dependency on @react-navigation/native
After updating to 2.0.4, I removed this patch: https://github.com/Iterable/react-native-sdk/issues/678#issue-3250445875

But now my build fails with

```
 Error: Unable to resolve module @react-navigation/native from /home/runner/work/app/node_modules/@iterable/react-native-sdk/lib/module/inbox/components/IterableInbox.js: @react-navigation/native could not be found within the project or in these directories:
  1 | "use strict";
  2 |
> 3 | import { useIsFocused } from '@react-navigation/native';
```

I'm not using the inbox feature or any UI component. Iterable should not depend on `@react-navigation/native`. We're using Wix's React-Native Navigation and do not have `@react-navigation/native` installed.

AdrienPoupa:
## ✏️ Description

Fixes https://github.com/Iterable/react-native-sdk/issues/770

This makes the `@react-navigation/native` lazy so users that do not depend on  or do not use any UI component can install the latest version of the RN SDK.

Full disclosure, this is mostly AI-generated and I was not able to test whether the `@react-navigation/native` import works since I'm not using it in my project.

### DIFF
--- a/src/core/hooks/index.ts
+++ b/src/core/hooks/index.ts
@@ -1,2 +1,3 @@
 export * from './useAppStateListener';
 export * from './useDeviceOrientation';
+export * from './useLazyIsFocused';
--- a/src/core/hooks/useLazyIsFocused.test.tsx
+++ b/src/core/hooks/useLazyIsFocused.test.tsx
@@ -1,0 +1,247 @@
+import { renderHook, act } from '@testing-library/react-native';
+import React from 'react';
+
+// Mock dynamic import behavior
+const mockUseIsFocused = jest.fn(() => true);
+
+// Mock the module before importing the hook
+jest.mock(
+  '@react-navigation/native',
+  () => ({
+    useIsFocused: mockUseIsFocused,
+  }),
+  {
+    virtual: true,
+  }
+);
+
+// Import after mocking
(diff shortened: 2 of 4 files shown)


AdrienPoupa:
A quick search yielded no other results, when excluding the examples and the comments (2 last results):

image

> Also, it should definitely alert the user if the module is being used and @react-navigation is not available. We use it for some important functionality.

Why? As far as I can tell, it's only used to find if the app is focused. I don't mind closing this if you prefer to handle the issue internally.

lposen:
> A quick search yielded no other results, when excluding the examples and the comments (2 last results):
> 
> image
> > Also, it should definitely alert the user if the module is being used and @react-navigation is not available. We use it for some important functionality.
> 
> Why? As far as I can tell, it's only used to find if the app is focused. I don't mind closing this if you prefer to handle the issue internally.

I'll keep it open for now.  I definitely would rather cut out any dependencies we can.  Unfortunately, this quarter is very tight, so I don't have the bandwidth to properly test this.  But as soon as I do get the bandwidth, I definitely want to.

PR #786, opened 2025-11-19
[SDK-196] Fix retryInterval not being able to be updated

lposen:
## 🔹 JIRA Ticket(s) if any

* [SDK-196](https://iterable.atlassian.net/browse/SDK-196)

## ✏️ Description

In Android, fix `retryInteraval` so that it will update correctly on re-initialization.

## Testing
1. cd into example 
2. In Terminal 1, run:
   		```bash
	 	watchman watch-del-all
		yarn start --reset-cache
		```
3. In terminal 2, run:
   		```bash
	 	yarn android
		```
4. In `useIterableApp.tsx`, uncomment line 210 (` return 'InvalidToken'`)
5. Open the console by pressing `j` in the terminal in which you ran `yarn start`
6. Click "Login" in the app.  
7. In the console, you should see "authHandler" being logged around every 5 seconds.
8. In `useIterableApp.tsx` find where `retryInterval` is set (~line 154). Update it to 2.
9. Refresh the app by pressing `r` in the terminal in which you ran `yarn start`
10. Click "Login" in the app.  
11. In the console, you should see "authHandler" being logged around every 2 seconds.

[SDK-196]: https://iterable.atlassian.net/browse/SDK-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/example/src/hooks/useIterableApp.tsx
+++ b/example/src/hooks/useIterableApp.tsx
@@ -151,7 +151,7 @@ export const IterableAppProvider: FunctionComponent<
 
       config.retryPolicy = {
         maxRetry: 5,
-        retryInterval:5,
+        retryInterval: 5,
         retryBackoff: IterableRetryBackoff.linear,
       };
 


android/src/main/java/com/iterable/reactnative/RNIterableAPIModuleImpl.java
```diff
@@ -88,7 +89,36 @@ public void initializeWithApiKey(String apiKey, ReadableMap configReadableMap, S
...
+        IterableConfig config = configBuilder.build();
+        IterableApi.initialize(reactContext, apiKey, config);
+
+        // Update retry policy on existing authManager if it was already created
+        // This fixes the issue where retryInterval is not respected after
+        // re-initialization
+        // TODO [SDK-197]: Fix the root cause of this issue, instead of this hack
+        try {
+            // Use reflection to access package-private fields and methods
+            java.lang.reflect.Field configRetryPolicyField = config.getClass().getDeclaredField("retryPolicy");
```
Ayyanchira: This looks risky!

Ayyanchira (approved):
Approving acknowledging that there is difference in how native layer RetryPolicy is created before initialization and when its created in ts layer. Im not sure exact problem it could lead to but feels like there could be cases where native layer and ts layer having their own retrypolicy running at different times causing some confusion.

Due to urgency, we are deciding to have this shipped. Something the team should address collectively as soon as bandwidth allows.

Also to note that addressing this later will involve docs update on when to configure RetryPolicy again. It could become a breaking change for developers.

PR #814, opened 2026-02-12
[SDK-248] notification-component

lposen:
## 🔹 JIRA Ticket(s) if any

* [SDK-248](https://iterable.atlassian.net/browse/SDK-248)

## ✏️ Description

Create notification component

## Testing
- Run example app
- Click on embedded tab
- Select "notification" as the view type
- Click "get messages for placement ids" -- the messages should show as notifications
- Click on the buttons -- **they should do whatever they were configured to do when you set the embedded message**
- Click "set view config" and try overriding the styles.  EG: set `{"backgroundColor":"blue"}`
	- **The style should update with your overrides**

https://github.com/user-attachments/assets/4108e6e4-0f55-4e23-8a8f-0a0d455c94cd

[SDK-248]: https://iterable.atlassian.net/browse/SDK-248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/embedded/components/IterableEmbeddedNotification/IterableEmbeddedNotification.styles.ts
+++ b/src/embedded/components/IterableEmbeddedNotification/IterableEmbeddedNotification.styles.ts
@@ -1,0 +1,54 @@
+import { StyleSheet } from 'react-native';
+
+export const styles = StyleSheet.create({
+  body: {
+    alignSelf: 'stretch',
+    fontSize: 14,
+    fontWeight: '400',
+    lineHeight: 20,
+  },
+  bodyContainer: {
+    display: 'flex',
+    flexDirection: 'column',
+    flexGrow: 1,
+    flexShrink: 1,
+    gap: 4,
+    width: '100%',
+  },
+  button: {
+    borderRadius: 32,
+    gap: 8,
+    paddingHorizontal: 12,
+    paddingVertical: 8,
+  },
+  buttonContainer: {
+    alignItems: 'flex-start',
+    alignSelf: 'stretch',
+    display: 'flex',
+    flexDirection: 'row',
+    gap: 12,
+    width: '100%',
+  },
+  buttonText: {
+    fontSize: 14,
+    fontWeight: '700',
+    lineHeight: 20,
+  },
+  container: {
+    alignItems: 'flex-start',
+    borderStyle: 'solid',
+    boxShadow:
+      '0 1px 1px 0 rgba(0, 0, 0, 0.06), 0 0 2px 0 rgba(0, 0, 0, 0.06), 0 0 1px 0 rgba(0, 0, 0, 0.08)',
+    display: 'flex',
+    flexDirection: 'column',
+    gap: 8,
+    justifyContent: 'center',
+    padding: 16,
+    width: '100%',
+  },
+  title: {
+    fontSize: 16,
+    fontWeight: '700',
+    lineHeight: 24,
+  },
+});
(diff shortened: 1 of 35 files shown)


src/embedded/components/IterableEmbeddedNotification/IterableEmbeddedNotification.tsx
```diff
@@ -0,0 +1,96 @@
...
+          styles.container,
+          {
+            backgroundColor: parsedStyles.backgroundColor,
+            borderColor: parsedStyles.borderColor,
+            borderRadius: parsedStyles.borderCornerRadius,
+            borderWidth: parsedStyles.borderWidth,
+          } as ViewStyle,
+        ]}
+      >
+        {}
```
joaodordio: this can be removed

src/embedded/hooks/useEmbeddedView/useEmbeddedView.ts
```diff
@@ -33,8 +34,10 @@ export const useEmbeddedView = (
   viewType: IterableEmbeddedViewType,
   /** The props for the embedded view. */
   {
-    config,
     message,
+    config,
+    onButtonClick = () => {},
+    onMessageClick = () => {},
```
joaodordio: the default params `onButtonClick = () => {}` and `onMessageClick = () => {}` create new function references on every render when the user doesn't pass them. 

This means `handleButtonClick` and `handleMessageClick` will recreate on every render despite the `useCallback`. 

There's also defaults on the Notification component for `onButtonClick` and `onMessageClick` before passing to the hook, so we are geting double-defaults and new refs. 

Should we use `useRef` for the callbacks (to avoid stale closures / unnecessary re-renders) or defaulting only in one place (component OR hook)?

src/embedded/hooks/useEmbeddedView/useEmbeddedView.ts
```diff
@@ -44,8 +47,28 @@ export const useEmbeddedView = (
     return getMedia(viewType, message);
   }, [viewType, message]);
 
+
```
joaodordio:
```suggestion
```

src/embedded/components/IterableEmbeddedNotification/IterableEmbeddedNotification.styles.ts
```diff
@@ -0,0 +1,54 @@
...
+  },
+  buttonText: {
+    fontSize: 14,
+    fontWeight: '700',
+    lineHeight: 20,
+  },
+  container: {
+    alignItems: 'flex-start',
+    borderStyle: 'solid',
+    boxShadow:
```
joaodordio, line 42: AI flagged this as boxShadow not working in earlier versions than RN 0.76.  

> _boxShadow as a string is a web/CSS property. In React Native, you'd typically use elevation (Android) and shadowColor/shadowOffset/shadowOpacity/shadowRadius (iOS). This will work in RN 0.76+ with the new architecture, but if you're targeting older versions it's a no-op._

joaodordio (approved):
LGTM just left some comments and some things AI flagged!

src/embedded/components/IterableEmbeddedNotification/IterableEmbeddedNotification.styles.ts
lposen: Good call!  Updating

src/embedded/hooks/useEmbeddedView/useEmbeddedView.ts
lposen: Good call.  Removed the defaults from the notification component and defaulted
to the hook.

src/embedded/hooks/useEmbeddedView/useEmbeddedView.ts
lposen: done

PR #834, opened 2026-04-06
[SDK-393] cleanup

lposen:
## 🔹 JIRA Ticket(s) if any

* [SDK-393](https://iterable.atlassian.net/browse/SDK-393)

## ✏️ Description

This is just a cleanup task.  Comments are cleaned up and documentation is added.

[SDK-393]: https://iterable.atlassian.net/browse/SDK-393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/embedded/components/IterableEmbeddedView.tsx
+++ b/src/embedded/components/IterableEmbeddedView.tsx
@@ -48,7 +48,6 @@ export interface IterableEmbeddedViewProps
  *   metadata: {
  *     messageId: 'test-message-123',
  *     placementId: 101,
- *     campaignId: 123456,
  *   },
  *   elements: {
  *     title: 'Test Title',
--- a/src/embedded/components/IterableEmbeddedView.tsx
+++ b/src/embedded/components/IterableEmbeddedView.tsx
@@ -9,7 +9,8 @@ import { IterableEmbeddedNotification } from './IterableEmbeddedNotification';
 /**
  * The props for the IterableEmbeddedView component.
  */
-export interface IterableEmbeddedViewProps extends IterableEmbeddedComponentProps {
+export interface IterableEmbeddedViewProps
+  extends IterableEmbeddedComponentProps {
   /** The type of view to render. */
   viewType: IterableEmbeddedViewType;
 }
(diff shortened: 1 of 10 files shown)


src/embedded/components/IterableEmbeddedView.tsx
```diff
@@ -20,10 +21,79 @@ interface IterableEmbeddedViewProps extends IterableEmbeddedComponentProps {
...
+ * const viewType = IterableEmbeddedViewType.Card;
+ *
+ * // Messages usually come from the embedded manager. `placementIds` is `number[] | null`
+ * // (use `null` to load messages for all placements), for example:
+ * // Iterable.embeddedManager.getMessages([101, 102]).then((messages) => { ... });
+ * const message: IterableEmbeddedMessage = {
+ *   metadata: {
+ *     messageId: 'test-message-123',
+ *     placementId: 101,
+ *     campaignId: 123456,
```
joaodordio: The type `IterableEmbeddedMessageMetadata` has `campaignId?: number | null`. The example implies it’s required. Minor, but could confuse developers.

src/embedded/components/IterableEmbeddedView.tsx
```diff
@@ -20,10 +21,79 @@ interface IterableEmbeddedViewProps extends IterableEmbeddedComponentProps {
...
+ *   },
+ *   elements: {
```
joaodordio, line 52: The type `IterableEmbeddedMessage` has `elements?: IterableEmbeddedMessageElements | null`. 

Same issue, a message with no elements is valid (e.g. a metadata-only message), but the example doesn’t reflect that.

joaodordio (approved):
Left too minor comments for
Improving clarity. But otherwise LGTM

src/embedded/components/IterableEmbeddedView.tsx
lposen: It's not exactly required for the view, but it would result in a very strange looking component as everything would be empty.

PR #854, opened 2026-04-07
fix: make @react-navigation/native an optional peer dependency

### LINKED ISSUE #770: Implicit dependency on @react-navigation/native

davidtruong:
Users who do not use the IterableInbox component (or who use a different navigation library) should not be forced to install @react-navigation/native.

- Mark @react-navigation/native, react-native-safe-area-context, and react-native-webview as optional in peerDependenciesMeta
- Create a safe useIsFocused hook wrapper that gracefully falls back to returning true when @react-navigation/native is not installed
- Update IterableInbox to use the wrapper instead of importing directly

Closes #770

https://claude.ai/code/session_01SsKoAbA48hJk6cDmbrRGqc

## 🔹 JIRA Ticket(s) if any

* [MOB-XXXX](https://iterable.atlassian.net/browse/MOB-XXXX)

## ✏️ Description

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,16 @@
     "react-native-webview": "*"
   },
   "peerDependenciesMeta": {
+    "@react-navigation/native": {
+      "optional": true
+    },
     "expo": {
+      "optional": true
+    },
+    "react-native-safe-area-context": {
+      "optional": true
+    },
+    "react-native-webview": {
       "optional": true
     }
   },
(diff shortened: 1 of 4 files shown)


davidtruong:
Things are fixed

PR #855, opened 2026-04-07
Fix createdAt and expiresAt returning numbers instead of Date objects

### LINKED ISSUE #581: IterableInAppMessage.createdAt type seems inconsistent with what's specified in the TS declaration - causing crashes
## Issue

The type is declared to be a `Date` (potentially undefined)

```ts
// https://github.com/Iterable/react-native-sdk/blob/master/ts/IterableInAppMessage.ts
declare class IterableInAppMessage {
    /**
     * the ID for the in-app message
     */
    readonly messageId: string;
    /**
     * the campaign ID for this message
     */
    readonly campaignId: number;
    /**
     * when to trigger this in-app
     */
    readonly trigger: IterableInAppTrigger;
    /**
     * when was this message created
     */
    readonly createdAt?: Date;
    /**
 ```
 
 However when i add the following logging:
 ```js
    console.log('=====================================')
    console.log(latestInboxMessage?.itbInAppMessage?.createdAt)
    console.log(typeof latestInboxMessage.itbInAppMessage.createdAt)
```

I see that `typeof createdAt` is a `string` when logging from Android and a Number when logging from iOS

Android logs:
_i included the error that is thrown that's causing my app to crash (android only)_
```
LOG  1720797936904
 LOG  string
 ERROR  Error: fromMillis requires a numerical input, but received a string with value 1720797936904
```

iOS Logs
```
LOG 1720797936904
 LOG  number
```

## Investigation (so far)
I'm not sure what is causing the typeof to be different between iOS/Android, however by reading the code it doesn't look like the typing of `Date | undefined` is correct. `dateObject.setUTCMilliseconds` returns a `Number`, not a `Date`

```js
// https://github.com/Iterable/react-native-sdk/blob/master/ts/IterableInAppMessage.ts#L115
    let createdAt = dict["createdAt"]
    if (createdAt) {
      var dateObject = new Date(0)
      createdAt = dateObject.setUTCMilliseconds(createdAt)
    }
```

davidtruong:
setUTCMilliseconds() returns a numeric timestamp, not a Date object, causing type mismatches and crashes on Android. Replace with new Date() constructor which correctly creates Date objects from millisecond timestamps. Also removes the downstream typeof workaround in defaultDateMapper and the @ts-ignore suppression.

Fixes #581

https://claude.ai/code/session_016EHrcqQDgexLB5xApCxrZj

## 🔹 JIRA Ticket(s) if any

* [MOB-XXXX](https://iterable.atlassian.net/browse/MOB-XXXX)

## ✏️ Description

### DIFF
--- a/src/inApp/classes/IterableInAppMessage.ts
+++ b/src/inApp/classes/IterableInAppMessage.ts
@@ -175,16 +175,8 @@ export class IterableInAppMessage {
     const campaignId = dict.campaignId;
     const trigger = IterableInAppTrigger.fromDict(dict.trigger);
 
-    let createdAt = dict.createdAt;
-    if (createdAt) {
-      const dateObject = new Date(0);
-      createdAt = dateObject.setUTCMilliseconds(createdAt);
-    }
-    let expiresAt = dict.expiresAt;
-    if (expiresAt) {
-      const dateObject = new Date(0);
-      expiresAt = dateObject.setUTCMilliseconds(expiresAt);
-    }
+    const createdAt = dict.createdAt ? new Date(dict.createdAt) : undefined;
+    const expiresAt = dict.expiresAt ? new Date(dict.expiresAt) : undefined;
     const saveToInbox = IterableUtil.readBoolean(dict, 'saveToInbox');
     const inboxMetadataDict = dict.inboxMetadata;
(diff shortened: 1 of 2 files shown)


davidtruong:
Things are fixed

PR #861, opened 2026-05-05
[SDK-475] update-rn-version-to-0-80

lposen:
## 🔹 JIRA Ticket(s) if any

* [SDK-475](https://iterable.atlassian.net/browse/SDK-475)

## ✏️ Description

[SDK-475]: https://iterable.atlassian.net/browse/SDK-475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/__mocks__/jest.setup.ts
```diff
@@ -3,7 +3,36 @@ import * as ReactNative from 'react-native';
 import { MockRNIterableAPI } from './MockRNIterableAPI';
 import { MockLinking } from './MockLinking';
 
-jest.mock('react-native/Libraries/EventEmitter/NativeEventEmitter.js');
+const mockNativeEventEmitter =
+  new (require('events').EventEmitter)() as import('events').EventEmitter;
+
+const mockNativeEventEmitterConstructor = jest.fn().mockImplementation(() => ({
```
joaodordio, line 9: The new mock replaces the per-instance auto-mock with a single module-scoped `EventEmitter` that all new `NativeEventEmitter(...)` instances share. That changes semantics: tests that subscribe to one "emitter" and emit on another will now incorrectly cross-talk, and listeners persist across tests since there's no reset. 

Two questions:
- Why not `mockImplementation(() => new (require('events').EventEmitter)()) `so each instance is isolated, like real RN behavior?
- If shared state is intentional, could we add a `beforeEach(() => mockNativeEventEmitter.removeAllListeners())` to prevent leakage between tests?

sumeruchat (approved):
LGTM for the 0.80.3 bump — local verification passed (install, typecheck, lint, tests, build, clean `pod install` + iOS example build under New Architecture). Android trusted to CI (no Java locally).

A few non-blocking items worth addressing before merge:

**1. iOS Pods cache key in CI is effectively broken (P2)**
`.github/workflows/ci.yml` keys the Pods cache on `hashFiles('example/ios/Podfile.lock')`, but `Podfile.lock` is gitignored, so that key is constant/empty across upgrades. On a cache hit, `pod install` is skipped — meaning a stale 0.79 Pods dir could survive into 0.80 CI runs. Reproduced locally: stale `fast_float 6.1.4` (0.79) vs RN 0.80.3's required `8.0.0`. After wiping Pods, clean `pod install` worked.

Suggested fixes (pick one):
- Key the cache on tracked inputs: `yarn.lock`, `package.json`, `example/ios/Podfile`.
- Always run `pod install` after restoring the cache (don't skip on hit).
- Long-term: commit `example/ios/Podfile.lock` so the cache key actually reflects pod state.

**2. README Legacy Architecture wording is ambiguous (P3)**
The new line in `README.md` says the SDK supports Legacy Arch but it "is no longer actively maintained. Use at your own risk." That can be read two ways — as RN's own 0.80 freeze, or as Iterable changing its support policy. Since the SDK still ships the legacy native path and `peerDependencies.react-native: "*"`, it'd be clearer to attribute the freeze to RN explicitly, e.g.:

> "Starting in React Native 0.80, RN's Legacy Architecture is frozen. This SDK still supports it for compatibility, but active development and testing is focused on the New Architecture."

If we're actually changing Iterable's support policy for legacy consumers, that belongs in release notes / version-support docs, not just a README warning.

**3. AGP pin (worth verifying in Android CI)**
`android/build.gradle:12` and `example/android/build.gradle:15` still pin AGP `8.7.2`, while RN 0.80's gradle plugin declares AGP `8.9.2` internally. Worth aligning with the RN 0.80 template if CI doesn't resolve it.

src/__mocks__/jest.setup.ts
lposen: It;s because in the real RN, the even emitter is actually shared.  So this is closer to the real behavior.

PR #866, opened 2026-05-20
[SDK-482] 0-84

lposen:
## 🔹 JIRA Ticket(s) if any

* [SDK-482](https://iterable.atlassian.net/browse/SDK-482)

## ✏️ Description

[SDK-482]: https://iterable.atlassian.net/browse/SDK-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/api/NativeRNIterableAPI.ts
```diff
@@ -122,12 +101,13 @@ export interface Spec extends TurboModule {
...
 
-  // Subscriptions
+  // Subscriptions (arrays only in spec — RN codegen rejects `T[] | null`; callers
+  // may still pass null at the bridge when using typed assertions.)
   updateSubscriptions(
-    emailListIds: number[] | null,
-    unsubscribedChannelIds: number[] | null,
-    unsubscribedMessageTypeIds: number[] | null,
-    subscribedMessageTypeIds: number[] | null,
+    emailListIds: number[],
```
joaodordio, line 107: The comment is honest about what's happening but the resulting contract is wrong. The TurboModule now declares `number[]` (non-null) while the public IterableApi.ts surface still accepts `number[] | null`, bridged by `as number[]` casts.

That cast silences TS but does nothing at runtime. The Iterable backend treats `null` as "do not change this field" and `[]` as "unsubscribe from everything" — those are not interchangeable. A consumer passing null today (per the documented contract) will, on the new architecture, end up unsubscribing the user from all channels of that category.

Fix should be one of:
1. Strip null at the JS layer (in IterableApi.ts) before calling RNIterableAPI. Make the public type `number[] | undefined` and short-circuit to a no-op when undefined.
2. Add an explicit null sentinel that the native side detects (e.g. `[-1]`) and converts back to nil on the Swift/Kotlin side.

Option 1 is cleaner. Either way, this needs:
- a Jest test that proves null does not become []
- a CHANGELOG entry calling out the public API change
- a paired docs PR

src/core/classes/IterableApi.ts
```diff
@@ -560,7 +560,7 @@ export class IterableApi {
     placementIds: number[] | null
   ): Promise<IterableEmbeddedMessage[]> {
     IterableLogger.log('getEmbeddedMessages: ', placementIds);
-    return RNIterableAPI.getEmbeddedMessages(placementIds);
+    return RNIterableAPI.getEmbeddedMessages(placementIds as number[]);
   }
```
joaodordio, line 564: These casts hide a real runtime mismatch (see comment on NativeRNIterableAPI.ts). Could we replace with explicit normalization, e.g.:
```
if (placementIds == null) {
  // backend semantics: omit field. Either skip the call or pass a sentinel.
  return Promise.resolve([]);
}
return RNIterableAPI.getEmbeddedMessages(placementIds);
```
and for updateSubscriptions, drop null inputs from the args object before calling the bridge.

example/android/build.gradle
```diff
@@ -12,7 +12,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath('com.android.tools.build:gradle:8.7.2')
+        classpath("com.android.tools.build:gradle")
```
joaodordio, line 16: The AGP version was stripped (`'com.android.tools.build:gradle:8.7.2'` → `"com.android.tools.build:gradle"`). This silently delegates the AGP version to the react-native-gradle-plugin's BOM. If that's intentional for RN 0.84, please add a Groovy comment explaining why, since:
- Gradle 9 may warn on classpath deps without coordinates
- the AGP version will now float with every RN patch
- consumers customizing buildscript dependency order can hit unexpected resolutions
If unintentional, pin to whatever AGP RN 0.84.1 ships with.

joaodordio (changes requested):
PR description is empty. For a change this big, please add:
- breaking changes: subscriptions/embedded null handling (with link to fix discussion above)
- bundle ID change 
- AGP version delegation (intentional or not)
- evidence of: example app smoke on iOS + Android, both archs; jest suite green; codegen output diff reviewed

example/android/build.gradle
lposen: Thanks for flagging this — the version removal is intentional.

This matches the RN 0.84 app template ([`react-native-community/template` `0.84-stable`](https://github.com/react-native-community/template/blob/0.84-stable/template/android/build.gradle)) and the `create-react-native-library` example pattern: the example app does not pin AGP/Kotlin/RNGP versions in `buildscript`.

Versions are resolved via `@react-native/gradle-plugin`, which is included as a composite build in `settings.gradle` (`pluginManagement { includeBuild(...) }` + `includeBuild(...)`). The plugin’s version catalog (`node_modules/@react-native/gradle-plugin/gradle/libs.versions.toml`) defines AGP **8.12.0** and Kotlin **2.1.20** for RN 0.84.1. That’s the setup Meta documents for versionless classpath deps (see also [facebook/react-native#48953](https://github.com/facebook/react-native/issues/48953)).

I added a comment above the classpath block in `example/android/build.gradle` so future readers know this is deliberate and shouldn’t re-pin without a reason.

Note: this applies to the **example app** only. The library’s `android/build.gradle` still pins AGP explicitly for standalone `android/` builds (no RNGP `includeBuild` there). I’m updating that pin from `8.7.2` → `8.12.0` (and the library Gradle wrapper to meet AGP 8.12’s minimum) in a separate change so it stays aligned with our RN 0.84.1 target without fighting the host app’s RNGP-managed versions.

src/api/NativeRNIterableAPI.ts
lposen: RN codegen couldn’t parse number[] | null (and related unions) -- the spec was narrowed to number[] so the upgrade could build.  Trying out available options now.

jferrao-itrbl:
Both tickets **Done**. Master is 0.85.3.

PR #869, opened 2026-05-21
SDK-478 resolve init promise immediately on iOS SDK, matching Android SDK

joaodordio:
## 🔹 JIRA Ticket(s) if any

* [SDK-478](https://iterable.atlassian.net/browse/SDK-478)

### What this changes

The iOS bridge for `Iterable.initialize` now resolves the JS promise immediately after calling the native SDK's synchronous initializer, matching the behavior the Android bridge has had since day one. The native iOS SDK is fully usable the moment `IterableAPI.initialize(...)` returns; nothing about JS-side correctness needs the in-app message fetch to complete before the promise resolves.

### Why

The promise contract was originally introduced in commit `4c357126` (Feb 2021) with the intent of letting JS callers `await Iterable.initialize()` to know that the SDK is ready. The implementation wired the promise to the iOS SDK's `IterableAPI.initialize2(callback:)` overload, which is documented in the iOS SDK as an internal staging/test endpoint override and whose callback fires when the first in-app messages fetch settles, not when the SDK is ready to use. This was a workable approximation in 2021 because the fetch typically settled in under 2s on healthy networks.

Two later changes made the gap between "SDK ready" and "first fetch settles" observable as a multi-second to multi-minute hang:

- iOS SDK 6.6.6 (PR #997) added a synchronous keychain migration in `start()` that runs on the calling thread, which for cross-platform bridges is the main thread (`DispatchQueue.main.async`).
- JWT-enabled customers exercise an authentication refresh path during init that can consume the full default retry budget (~60s).

Prior consolidation attempts include commit `c07b5ccd` (Oct 2025) which made exactly the change this PR makes but did not survive the merge to master, and commit `f37cc10b` (April 2026) which added a 10s timeout band-aid as part of SDK-392 but also did not survive to master.

This PR consolidates the correction: bridge calls sync `IterableAPI.initialize(...)` and resolves `true` immediately. Behavior now matches Android's `RNIterableAPIModuleImpl.initializeWithApiKey` which has done the same since its introduction. The internal `initialize2` (endpoint override) path is preserved for staging/test callers but its callback is no longer waited on.

### Compatibility

Documented usage in `iterable-docs` already shows `Iterable.initialize('<KEY>', config);` without `await`, aligned with the new behavior. Any caller relying on the old wait semantics (the promise resolving only after the in-app fetch attempt) would have been exposed to an unbounded hang under SDK-478 conditions, so the likelihood of customer code depending on it is low. The behavior change is called out explicitly in the CHANGELOG and the release notes.

### Tests

New SDK-478 contract-pin tests in `src/core/classes/IterableApi.test.ts`: `initializeWithApiKey` and `initialize2WithApiKey` both assert that the JS-side call resolves in under 50ms when the native mock resolves true, pinning the "pure passthrough, no additional await" contract.

### Linked tickets

- [SDK-392](https://iterable.atlassian.net/browse/SDK-392) (the customer-facing hang report)
- [SDK-478](https://iterable.atlassian.net/browse/SDK-478) (validation work that surfaced the bridge contract issue)

[SDK-478]: https://iterable.atlassian.net/browse/SDK-478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-392]: https://iterable.atlassian.net/browse/SDK-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joaodordio:
> [!NOTE]
> Heads up to reviewers: the failing `Analyze (java-kotlin)` check on the CodeQL Advanced workflow is **pre-existing on `master`** and unrelated to this PR.

**Evidence.** The job has been red on every `master` push since `55ac386b` (merge of #860, RN 0.79 upgrade, 2026-05-15). Latest `master` HEAD (`de793ec`, 2026-05-20) still fails the same way. The last successful run on master was `e69c305c` on 2026-03-24, before the RN 0.79 work landed.

**Root cause.** [`example/android/settings.gradle:3`](https://github.com/Iterable/react-native-sdk/blob/master/example/android/settings.gradle#L3) calls `autolinkLibrariesFromCommand()` (the RN 0.79+ autolinking entry point), which shells out to `npx @react-native-community/cli config` from inside Gradle's settings evaluation. The Gradle process Run by CodeQL's build tracer doesn't inherit `npx` on its PATH, so the subprocess returns exit 127 ("command not found") and the build aborts.

**Why it can't be from this PR.** The diff is iOS-only - no Android files, no Gradle files, no workflow files touched. Reproduces verbatim on `master` without any of this PR's changes.

**Suggested follow-up (separate PR).** Either add an explicit `actions/setup-node` step right before the `./gradlew clean` step in [`.github/workflows/codeql.yml`](https://github.com/Iterable/react-native-sdk/blob/master/.github/workflows/codeql.yml), or swap the `npx` call in `settings.gradle` for the explicit node + cli path overload:

```groovy
ex.autolinkLibrariesFromCommand([
  "node",
  "node_modules/@react-native-community/cli/build/bin.js",
  "config"
])

File: ios/RNIterableAPI/ReactIterableAPI.swift [hunk 1]
```diff
@@ -645,24 +645,40 @@ import React
...
     DispatchQueue.main.async {
-      IterableAPI.initialize2(
-        apiKey: apiKey,
-        launchOptions: launchOptions,
-        config: iterableConfig,
-        apiEndPointOverride: apiEndPointOverride
-      ) { result in
-        resolver(result)
+      // The native iOS SDK is fully usable the moment IterableAPI.initialize
+      // returns. The legacy initialize2(callback:) overload fires its callback
```
lposen, line 672: From what I can tell, we're not waiting for `Iterable.initialize2` to return, though.  It looks like we're bypassing any return and always resolving as true.  Would this not silence any errors that may occur?

joaodordio: Good question, but the callback was never an init-error signal. Two pieces of evidence from the iOS SDK source.

**1. `IterableAPI.initialize(...)` and `initialize2(...)` run the same code, neither is failable.**

[`swift-sdk/SDK/IterableAPI.swift#L102-L108`](https://github.com/Iterable/iterable-swift-sdk/blob/master/swift-sdk/SDK/IterableAPI.swift#L102-L108):

```swift
public static func initialize(apiKey: String,
                              launchOptions: [UIApplication.LaunchOptionsKey: Any]?,
                              config: IterableConfig = IterableConfig()) {
    initialize2(apiKey: apiKey,
                launchOptions: launchOptions,
                config: config)
}
```

The public sync `initialize(...)` is just `initialize2(...)` with no callback. Both return `Void`, no `throws`, no `Result`. There is no init-error channel on the native API.

**2. The Bool the old bridge was awaiting represented the first in-app fetch, not init.**

[`swift-sdk/SDK/IterableAPI.swift#L113-L128`](https://github.com/Iterable/iterable-swift-sdk/blob/master/swift-sdk/SDK/IterableAPI.swift#L113-L128):

```swift
public static func initialize2(apiKey: String,
                               launchOptions: [UIApplication.LaunchOptionsKey: Any]?,
                               config: IterableConfig = IterableConfig(),
                               apiEndPointOverride: String? = nil,
                               callback: ((Bool) -> Void)? = nil) {
    AppExtensionHelper.initialize()
    implementation = InternalIterableAPI(apiKey: apiKey, ...)
    _ = implementation?.start().onSuccess { _ in
        callback?(true)
    }.onError { _ in
        callback?(false)
    }
    ...
}
```

`start()` is [`InAppManager.start()`](https://github.com/Iterable/iterable-swift-sdk/blob/master/swift-sdk/Internal/in-app/InAppManager.swift#L186-L196), which does `return scheduleSync()`, the first in-app messages fetch. So the old `false` only ever meant "first in-app fetch failed", which is exactly the signal SDK-392 wants us to stop blocking on.

**3. Android is already on this contract.**

[`RNIterableAPIModuleImpl.java#L139-L141`](https://github.com/Iterable/react-native-sdk/blob/master/android/src/main/java/com/iterable/reactnative/RNIterableAPIModuleImpl.java#L139-L141):

```java
// MOB-10421: Figure out what the error cases are and handle them appropriately
// This is just here to match the TS types and let the JS thread know when we are done initializing
promise.resolve(true);
```

Android resolves `true` immediately after sync `IterableApi.initialize(...)` and the comment explicitly acknowledges there are no error cases to surface. This PR brings iOS to that same contract.

**Net:** nothing real is being silenced. If a downstream caller cares about the in-app fetch result (the only signal the old callback carried), it is still observable via `inAppManager` listeners.

lposen (commented):
I'm still apprehensive about this fix, but approved assuming that a comment will be put in explainint it, and tying it to a specific ticket to fix error bubbling

joaodordio:
> I'm still apprehensive about this fix, but approved assuming that a comment will be put in explainint it, and tying it to a specific ticket to fix error bubbling

[Epic](https://iterable.atlassian.net/browse/SDK-532) created!

jferrao-itrbl (approved):

PR #877, opened 2026-06-19
[SDK-363] Expose register device token

jferrao-itrbl:
## 🔹 JIRA Ticket(s) if any

* [SDK-363](https://iterable.atlassian.net/browse/SDK-363)

## ✏️ Description

Expose `registerDeviceToken` in React Native bridge:
   - Add `Iterable.registerDeviceToken(token)` public API
   - Add bridge proxy `IterableApi.registerDeviceToken(token)`
   - Add TurboModule spec entry in `NativeRNIterableAPI.ts`
   - Add iOS Swift/Obj-C++ glue with hex-to-Data conversion for APNS tokens
   - Add Android impl + old/new architecture stubs for direct FCM token pass-through
   - Update `MockRNIterableAPI` and unit tests
   - Add CHANGELOG entry under Unreleased

[SDK-363]: https://iterable.atlassian.net/browse/SDK-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ios/RNIterableAPI/ReactIterableAPI.swift
+++ b/ios/RNIterableAPI/ReactIterableAPI.swift
@@ -610,7 +610,8 @@ import React
   private let inboxSessionManager = InboxSessionManager()
 
   private func data(fromHex hex: String) -> Data? {
-    var data = Data()
+    guard !hex.isEmpty, hex.count.isMultiple(of: 2) else { return nil }
+    var data = Data(capacity: hex.count / 2)
     var chars = hex.makeIterator()
     while let high = chars.next(), let low = chars.next() {
       guard let highValue = high.hexDigitValue, let lowValue = low.hexDigitValue else {
--- a/ios/RNIterableAPI/ReactIterableAPI.swift
+++ b/ios/RNIterableAPI/ReactIterableAPI.swift
@@ -144,6 +144,16 @@ import React
     IterableAPI.disableDeviceForCurrentUser()
   }
 
+  @objc(registerDeviceToken:)
+  public func registerDeviceToken(token: String) {
+    ITBInfo()
+    guard let tokenData = data(fromHex: token) else {
+      ITBError("Could not convert token to Data: invalid hex string")
+      return
+    }
+    IterableAPI.register(token: tokenData)
+  }
+
   @objc(getLastPushPayload:rejecter:)
   public func getLastPushPayload(resolver: RCTPromiseResolveBlock, rejecter: RCTPromiseRejectBlock)
   {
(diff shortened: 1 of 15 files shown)


ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -599,6 +609,18 @@ import React
 
   private let inboxSessionManager = InboxSessionManager()
 
+  private func data(fromHex hex: String) -> Data? {
```
joaodordio, line 612: data(fromHex:) returns a non-nil empty Data for empty input, so registerDeviceToken("") falls through and calls IterableAPI.register(token: Data()). That registers an empty APNS token with no error surfaced. Add an hex.isEmpty guard at the top of data(fromHex:) (or in registerDeviceToken before the call) and log the same ITBError so callers see something in their logs.

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -144,6 +144,16 @@ import React
     IterableAPI.disableDeviceForCurrentUser()
   }
 
+  @objc(registerDeviceToken:)
+  public func registerDeviceToken(token: String) {
+    ITBInfo()
+    guard let tokenData = data(fromHex: token) else {
```
joaodordio, line 150: if the token is empty, data(fromHex:) will not fail here because it returns an empty Data().

ios/RNIterableAPI/ReactIterableAPI.swift
```diff
@@ -599,6 +609,18 @@ import React
 
   private let inboxSessionManager = InboxSessionManager()
 
+  private func data(fromHex hex: String) -> Data? {
+    var data = Data()
+    var chars = hex.makeIterator()
+    while let high = chars.next(), let low = chars.next() {
+      guard let highValue = high.hexDigitValue, let lowValue = low.hexDigitValue else {
```
joaodordio, line 617: NIT: Odd-length input gets silently truncated. The last hex char is consumed by high and the loop exits on low == nil, so we return a token shorter than the source with no error. APNS tokens are always even length, but customers paste tokens from unpredictable places and a malformed input should fail loudly. Reject inputs whose count isn't a multiple of 2 (return nil) so the existing ITBError branch fires.

joaodordio (changes requested):
Left some comments on the data(fromHex:) method that could cause regressions, but those are simple changes!

Suggestion:
``` swift
private func data(fromHex hex: String) -> Data? {
  guard !hex.isEmpty, hex.count.isMultiple(of: 2) else { return nil }
  var data = Data(capacity: hex.count / 2)
  var chars = hex.makeIterator()
  while let high = chars.next(), let low = chars.next() {
    guard let highValue = high.hexDigitValue, let lowValue = low.hexDigitValue else {
      return nil
    }
    data.append(UInt8(highValue << 4 | lowValue))
  }
  return data
}
```
Everything else looks good! Nice work!

lposen (approved):

joaodordio (approved):

PR #880, opened 2026-06-30
[SDK-518] - Increase test coverage

jferrao-itrbl:
## 🔹 JIRA Ticket(s) if any

* [SDK-518](https://iterable.atlassian.net/browse/SDK-518)

## ✏️ Description

- Increate unit test coverage on core SDK surface.
- Add mechanical coverage ratchet for models, enums and utils.

Aggregate coverage: 71.07% statements | 60.55% branch | 66.66% functions | 71.14% lines

[SDK-518]: https://iterable.atlassian.net/browse/SDK-518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/embedded/classes/IterableEmbeddedManager.test.ts
+++ b/src/embedded/classes/IterableEmbeddedManager.test.ts
@@ -657,6 +657,36 @@ describe('IterableEmbeddedManager', () => {
       // THEN isEnabled should be false (default)
       expect(manager.isEnabled).toBe(false);
     });
+
+    it('should initialize with embedded messaging disabled when config flag is null (nullish coalescing fallback)', () => {
+      // GIVEN a config where enableEmbeddedMessaging is explicitly null
+      // (exercises the `?? false` branch on line 57)
+      const configWithNull = new IterableConfig();
+      (configWithNull as unknown as {
+        enableEmbeddedMessaging: boolean | null;
+      }).enableEmbeddedMessaging = null;
+
+      // WHEN creating a new embedded manager
+      const manager = new IterableEmbeddedManager(configWithNull);
+
+      // THEN isEnabled should fall back to false via the `?? false` branch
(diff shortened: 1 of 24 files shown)


src/embedded/classes/IterableEmbeddedManager.test.ts
```diff
@@ -657,6 +657,36 @@ describe('IterableEmbeddedManager', () => {
...
+      }).enableEmbeddedMessaging = null;
+
+      // WHEN creating a new embedded manager
+      const manager = new IterableEmbeddedManager(configWithNull);
+
+      // THEN isEnabled should fall back to false via the `?? false` branch
+      expect(manager.isEnabled).toBe(false);
+    });
+
+    it('should keep isEnabled false when constructed with enableEmbeddedMessaging === false and not call native sync', () => {
```
joaodordio: Nit: the test title `...and not call native sync` reads at first glance like the assertion should be `not.toHaveBeenCalled()`. The body and `toHaveBeenCalledTimes(1)` actually prove "constructor does not auto-call native sync, explicit `syncMessages()` still proxies through", which is fine, but the name buries that.

Consider either:

- Renaming to something like `constructor with enableEmbeddedMessaging=false does not auto-call native sync, and explicit syncMessages still proxies through`, or
- Splitting into two tests, one for the constructor invariant and one for the proxy behavior.

Optional bonus: add an explicit `expect(MockRNIterableAPI.syncEmbeddedMessages).not.toHaveBeenCalled()` between `new IterableEmbeddedManager(...)` and `manager.syncMessages()` to pin down the constructor-no-auto-sync invariant independently of the proxy call count.

joaodordio (approved):

PR #882, opened 2026-07-02
feat(SDK-520): replace magic 1000ms timeouts with named, configurable…

jferrao-itrbl:
## 🔹 JIRA Ticket(s) if any

* [SDK-520](https://iterable.atlassian.net/browse/SDK-520)

## ✏️ Description

Replace hardcoded magic timeouts:
- Add IterableConfig.androidWakeDelayMs and authCallbackTimeoutMs to expose the two hardcoded timeouts as tunable, documented values.
- Replace the fixed-window auth callback gate with an event-driven Promise latch; the timer survives only as a safety-net fallback.

[SDK-520]: https://iterable.atlassian.net/browse/SDK-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/classes/Iterable.ts
+++ b/src/core/classes/Iterable.ts
@@ -28,6 +28,20 @@ const RNEventEmitter = new NativeEventEmitter(RNIterableAPI);
 const defaultConfig = new IterableConfig();
 
 /**
+ * Fallback safety-net timeout for the auth callback latch when no native
+ * auth success/failure event arrives. Overridable via
+ * {@link IterableConfig.authCallbackTimeoutMs}.
+ */
+const AUTH_CALLBACK_TIMEOUT_DEFAULT_MS = 1000;
+
+/**
+ * Default delay (ms) the SDK waits on Android before invoking the URL handler
+ * so the host Activity can wake from the background. Overridable via
+ * {@link IterableConfig.androidWakeDelayMs}.
+ */
+const ANDROID_WAKE_DELAY_DEFAULT_MS = 1000;
+
+/**
  * Checks if the response is an IterableAuthResponse
  */
 const isIterableAuthResponse = (
@@ -1009,10 +1023,20 @@ export class Iterable {
         Iterable.wakeApp();
 
         if (Platform.OS === 'android') {
-          //Give enough time for Activity to wake up.
-          setTimeout(() => {
+          // Give the host Activity time to wake from the background before
+          // dispatching the URL to the handler. Without this delay the
+          // handler can race the activity lifecycle and drop the link on
+          // cold start. Tunable via IterableConfig.androidWakeDelayMs.
+          const wakeDelayMs =
+            Iterable.savedConfig.androidWakeDelayMs ??
+            ANDROID_WAKE_DELAY_DEFAULT_MS;
+          if (wakeDelayMs > 0) {
+            setTimeout(() => {
+              callUrlHandler(Iterable.savedConfig, url, context);
+            }, wakeDelayMs);
+          } else {
             callUrlHandler(Iterable.savedConfig, url, context);
-          }, 1000);
+          }
         } else {
           callUrlHandler(Iterable.savedConfig, url, context);
         }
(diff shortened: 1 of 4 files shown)


src/core/classes/Iterable.ts
```diff
@@ -1043,37 +1067,90 @@ export class Iterable {
...
+                    // well.
                     promiseResult.failureCallback?.();
+                  } else {
+                    IterableLogger?.log('No callback received from native layer');
                   }
-                } else {
-                  IterableLogger?.log('No callback received from native layer');
                 }
-              }, 1000);
+              );
```
joaodordio, line 1178: `authLatchResolver` and `pendingAuthResult` are shared across every `handleAuthCalled`, so overlapping auth invocations can wipe each other's state.

Concrete failure:
1. Auth # 1: `handleAuthCalled` resets, `authHandler` resolves, latch # 1 created, `authLatchResolver = resolve1`, race # 1 starts.
2. Auth # 2: `handleAuthCalled` fires while # 1 is still racing. Reset nulls `authLatchResolver`, `authHandler` # 2 resolves, latch # 2 created, `authLatchResolver = resolve2`.
3. Race # 1's safety-net timer fires, resolves race # 1 with `NO_CALLBACK`. The `.then` handler runs `authLatchResolver = null` and `pendingAuthResult = null`, wiping resolve2 and any buffered result belonging to # 2.
4. Native `handleAuthSuccessCalled` for either invocation now has nowhere to land. Native completes the request, JS silently drops the `successCallback` / `failureCallback`.

In production this looks like JWT refresh flows that succeed on the server but the app never learns about it.

Please scope this state per invocation, e.g. capture the resolver/buffer inside a per-call object and only null out that object's fields in its own `.then`. A test that fires two `handleAuthCalled` events back to back and asserts both `successCallback`s are invoked would lock this down.

src/core/classes/Iterable.ts
```diff
@@ -27,6 +27,20 @@ const RNEventEmitter = new NativeEventEmitter(RNIterableAPI);
 
 const defaultConfig = new IterableConfig();
 
+/**
+ * Fallback safety-net timeout for the auth callback latch when no native
+ * auth success/failure event arrives. Overridable via
+ * {@link IterableConfig.authCallbackTimeoutMs}.
+ */
+const AUTH_CALLBACK_TIMEOUT_DEFAULT_MS = 1000;
```
joaodordio: Default of 1000ms defeats the "safety net" framing.

Native success/failure only fires after `passAlongAuthToken` and a full HTTP round-trip against Iterable, which routinely exceeds 1s on real mobile networks. With this default, the timer wins the race first, the resolver-clearing branch runs, and the late native event ends up hitting the null-resolver path so `successCallback` / `failureCallback` gets dropped.

Native waits up to 30s for us on both platforms (`CountDownLatch.await(30, TimeUnit.SECONDS)` on Android, `authHandlerSemaphore.wait(timeout: 30.0)` on iOS), so JS closing the loop at 1s is asymmetric.

My suggestion would be bumping the default to 5000-10000ms so the timer is genuinely a fallback for stuck native, not a race the timer usually wins.

joaodordio (changes requested):

joaodordio:
Really nice iteration on the latch, the FIFO scoping and the 6s default both look great. One thing I noticed while tracing through the new design that I think is worth catching before this lands:

The FIFO queue only gets drained in the `isIterableAuthResponse` branch and in the two native listeners. The `.catch` and the three other `.then` branches (bare string, null/undefined, unexpected type) push an `AuthInvocation` but never remove it, so it sits at the head of `pendingAuthInvocations` as a zombie. Since native emits `handleAuthSuccessCalled` or `handleAuthFailureCalled` in all of those paths too (`passAlongAuthToken(null)` → `handleAuthTokenSuccess(null)` → `handleAuthFailureCalled` on Android, semaphore always emits on iOS, and the `.catch` path hits the 30s native latch timeout which still returns and emits), the zombie ends up catching that event and buffering it into `pendingResult` where nothing consumes it.

Concrete single-flight trace for the `.catch` case, which is probably the most common non-happy path in production:

1. `handleAuthCalled` # 1 fires, `inv1` pushed. Queue `[inv1]`.
2. Customer's `authHandler` throws (network blip, exception in their token fetch, etc). `.catch` logs. `inv1` stays.
3. Native latch times out at 30s, emits `handleAuthFailureCalled` (Android returns null token → `AUTH_TOKEN_NULL`; iOS semaphore timeout).
4. Listener finds `inv1` at head with no resolver, buffers `inv1.pendingResult = FAILURE`. `onJwtError` still fires correctly, all good there.
5. SDK schedules a retry via `IterableAuthManager.scheduleAuthTokenRefresh`, eventually emits `handleAuthCalled` # 2.
6. `inv2` pushed. Queue `[inv1, inv2]`.
7. `authHandler` # 2 succeeds with an `IterableAuthResponse`, latch wired on `inv2`.
8. Native # 2 succeeds, emits `handleAuthSuccessCalled`.
9. Listener routes to `queue[0]` which is still `inv1` → buffers `SUCCESS` on the zombie.
10. `inv2`'s latch never sees it, safety-net wins at 6s, "No callback received" logged, `successCallback2` silently dropped.

Also a slow memory leak on the `.catch` path since every rejected `authHandler` adds one `AuthInvocation` for the process lifetime.

The fix I'd suggest is dropping the invocation from the queue in each of those four paths, same shape as the defensive splice you already have in `race.then`, something like:

```ts
const removeInvocation = () => {
  const index = pendingAuthInvocations.indexOf(invocation);
  if (index !== -1) {
    pendingAuthInvocations.splice(index, 1);
  }
};
```
then call removeInvocation() in the string, null/undefined, unexpected-type branches, and in the .catch. There's a tiny caveat that a late native event for a .catched invocation will then misroute to whatever is at the new head (either nothing, or a newer invocation), but that's a pre-existing bridge limitation with no correlation id and the current behavior is strictly worse since it deterministically blocks the next real callback.

Locking this down with a rejection-then-retry test in the same shape as the great overlap regression test you just added would be perfect, something like: fire handleAuthCalled → reject authHandler → fire handleAuthCalled again → resolve with IterableAuthResponse → assert the second successCallback fires.

jferrao-itrbl:
> 

Thanks for catching this @joaodordio , really good spot.

Added a `removeInvocation()` helper scoped to each invocation and called it in all four non-happy paths, dropping the invocation from the queue before native's late event arrives. The no-correlation-id bridge limitation (a late native event for a removed invocation may misroute to the new queue head) is noted in a comment. Strictly better than the zombie-at-head block; full overlap hardening with a correlation id would be a separate bridge-contract change.

Added a regression test: `handleAuthCalled` → reject `authHandler` → `handleAuthCalled` again → resolve with `IterableAuthResponse` → emit native success → assert the second `successCallback` fires and no "No callback received" warning is logged.

All tests green and updated rules to make sure similar scenarios are avoided or caught earlier in the future.

joaodordio (approved):
Thanks for addressing the issues João!


LGTM!

PR #884, opened 2026-07-16
[SDK-118] Upgrade eslint

jferrao-itrbl:
## 🔹 JIRA Ticket(s) if any

* [SDK-118](https://iterable.atlassian.net/browse/SDK-118)

## ✏️ Description

- Upgrade ESLint to v9
- Resolve eqeqeq lint warnings in IterableConfig

[SDK-118]: https://iterable.atlassian.net/browse/SDK-118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -87,11 +87,11 @@
     "baseline-browser-mapping": "^2.10.29",
     "commitlint": "^19.6.1",
     "del-cli": "^5.1.0",
-    "eslint": "^8.51.0",
+    "eslint": "^9.39.5",
     "eslint-config-prettier": "^10.1.8",
```
rtlsilva, line 93: If I understand correctly, the now-removed `.eslintrc.js` extended `prettier` to disable ESLint rules that conflicted with it, and that was the only place where it was referenced.
The new flat config drops that reference entirely, so is this now a dead dependency, or something that should be re-added to the flat config?
Either way, this is a non-blocking observation.

rtlsilva (approved):

package.json
jferrao-itrbl: Good catch worth verifying.

Not dead: `eslint-config-prettier` is applied transitively via `@react-native/eslint-config/flat` (it requires and includes it as the first config block, pulled in through `...reactNativeConfig`). The old `.eslintrc.js` behaviour is preserved. No re-add needed; the package.json pin is just kept as a direct dep.

PR #890, opened 2026-07-20
[SDK-122]: Upgrade react-native-builder-bob to 0.43

jferrao-itrbl:
Align bob across workspaces and migrate Metro to
react-native-monorepo-config with the iterable-react-native-sdk-source export condition. Also set CI=true for the circular lefthook command (same as SDK-121) until that fix lands on master.

## 🔹 JIRA Ticket(s) if any

* [SDK-122](https://iterable.atlassian.net/browse/SDK-122)

## ✏️ Description

Align bob across workspaces and migrate Metro to `react-native-monorepo-config` with the `iterable-react-native-sdk-source` export condition. Also set `CI=true` for the circular lefthook command (same as SDK-121) until that fix lands on master.

[SDK-122]: https://iterable.atlassian.net/browse/SDK-122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -9,6 +9,8 @@ pre-commit:
       run: npx tsc
     circular:
       glob: "src/**/*.{ts,tsx}"
+      env:
+        CI: "true"
       run: yarn check:circular
     # version-check:
     #   glob: "package.json"
--- a/example/metro.config.js
+++ b/example/metro.config.js
@@ -1,9 +1,6 @@
 const path = require('path');
 const { getDefaultConfig } = require('@react-native/metro-config');
-const { getConfig } = require('react-native-builder-bob/metro-config');
-const pkg = require('../package.json');
-
-const root = path.resolve(__dirname, '..');
+const { withMetroConfig } = require('react-native-monorepo-config');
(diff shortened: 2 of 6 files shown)


lefthook.yml
```diff
@@ -9,6 +9,8 @@ pre-commit:
       run: npx tsc
     circular:
       glob: "src/**/*.{ts,tsx}"
+      env:
+        CI: "true"
```
rtlsilva, line 13: The PR description mentions this is a temporary workaround until a specific fix lands. Should an inline comment be added for future reference?

rtlsilva (approved):
Nothing major to comment here, just reading the diff but seems like a clean and well-scoped tooling upgrade. LGTM

lefthook.yml
jferrao-itrbl: Should be aligned as soon as updated with the latest `master` anyway so cleaner as soon as merged.

PR #900, opened 2026-09-15
docs(SDK-549): document logout() as native logoutUser equivalent

jferrao-itrbl:
# 📝 Summary
Document `Iterable.logout()` as the RN equivalent of native `logoutUser`; no runtime or bridge change.

## 📖 Description
RN already ships `Iterable.logout()`, which removes JS listeners and clears identity via `setEmail(null)` + `setUserId(null)`. That identity-clear path is the native logout cleanup on both platforms (auth reset, in-app/embedded reset, push disable when `autoPushRegistration` is on). Android has no public `logoutUser`.

This PR only updates the `logout()` JSDoc so that mapping is explicit, including:
- why both identity-clears are required (iOS no-ops the identifier that is not set)
- that iOS `logoutUser(withOnSuccess:onFailure:)` is not exposed
- that calling `logout()` with no signed-in user is safe

No new TurboModule method, no codegen, no behavior change.

## 🧪 How to test?
- [x] `yarn typecheck`
- [x] `yarn test` (existing `logout` tests in `src/core/classes/Iterable.test.ts`)
- [x] Confirm `Iterable.logout()` still calls `setEmail(null)` and `setUserId(null)`
- [x] Optional: in the example app, sign in then log out; identity is cleared and the app does not crash when logging out with no user

## 🧾 Changelog
No customer facing changes

## 📹 Loom recording if applicable
N/A

#### 🐞 Github Issues solved
N/A

#### 📚 Docs PR if applicable
N/A — JSDoc only; no iterable-docs change.

[SDK-549]: https://iterable.atlassian.net/browse/SDK-549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core/classes/Iterable.ts
```diff
@@ -941,7 +941,27 @@ export class Iterable {
...
+   * - disabling the current push token when
+   *   `IterableConfig.autoPushRegistration` is `true`
+   *
+   * Both identity-clears are required. On iOS, `setEmail(null)` is a no-op
+   * when the user is identified only by userId, and `setUserId(null)` is a
+   * no-op when the user is identified only by email. Android has no public
+   * `logoutUser`; private `logoutPreviousUser()` runs as part of identity
+   * clear.
+   *
+   * iOS `logoutUser(withOnSuccess:onFailure:)` is not exposed.
```
franco-zalamena-iterable, line 961: From what i checked iOS does have a public logout method, it is just not exposed in the bridge, android actually doesn't have and maybe that's something for us to create a followup ticket to fix.

But i think we can rephrase this to make it more accurate and think of making it exposed through the bridge when we get both platforms

franco-zalamena-iterable (changes requested):
Just to double check some points

src/core/classes/Iterable.ts
jferrao-itrbl: +1 on having iOS and Android aligned by adding this on Android.

iOS does have public `logoutUser()`; RN just does not bridge it. Android still has no public API (`logoutPreviousUser()` is private). I'll rephrase the JSDoc to say that. Bridging a dedicated `logoutUser` waits until Android has a public method so the RN surface is the same on both platforms.

franco-zalamena-iterable (approved):

PR #901, opened 2026-09-16
feat(SDK-550): expose disableDeviceForAllUsers on the RN SDK

jferrao-itrbl:
# 📝 Summary
Add `Iterable.disableDeviceForAllUsers()` so RN apps can unregister this device's push token for every associated user on iOS.

## 📖 Description
The RN SDK only bridged `disableDeviceForCurrentUser`. iOS already has public `IterableAPI.disableDeviceForAllUsers()`; Android does not have a public "all users" equivalent (`disablePush()` is current-user only; `disableToken(...)` is protected).

This PR adds `Iterable.disableDeviceForAllUsers()` / `IterableApi.disableDeviceForAllUsers()` (no args, fire-and-forget):

- **iOS:** forwards to `IterableAPI.disableDeviceForAllUsers()`.
- **Android:** logged no-op (`IterableLogger.w`); does not throw. Use `disableDeviceForCurrentUser()` for current-user disable.

`disableDeviceForCurrentUser` is unchanged. Native Android public API and iOS `async throws` / callback variants are out of scope.

## 🧪 How to test?
- [ ] `yarn typecheck` and `yarn test` pass.
- **iOS:** initialize, register a token, call `Iterable.disableDeviceForAllUsers()`, confirm the device token is disabled for all users associated with the device (same native behavior as `IterableAPI.disableDeviceForAllUsers()`).
- **iOS edge:** call before `initialize()` / with no token — should not crash (native handles internally; RN does not surface success/failure).
- **Android:** call `Iterable.disableDeviceForAllUsers()` — warning in logcat, no crash; `disableDeviceForCurrentUser()` still disables push for the current user.
- After `NativeRNIterableAPI.ts` change, regenerate codegen (`cd example/ios && pod install`). Note: full `pod install` may currently fail on the pre-existing `Iterable-iOS-SDK` pin mismatch (podspec `6.6.7` vs `Podfile.lock` `6.7.5`, SDK-709). Codegen for this method already succeeds.

## 🧾 Changelog
Unreleased: added `Iterable.disableDeviceForAllUsers()` (iOS native forward; Android logged no-op).

## 📹 Loom recording if applicable
N/A

#### 🐞 Github Issues solved
N/A

#### 📚 Docs PR if applicable
JSDoc on `Iterable.disableDeviceForAllUsers()` documents the iOS-only effect and Android no-op (picked up by TypeDoc). Follow-up docs-repo PR if product docs should mention the new method.

[SDK-550]: https://iterable.atlassian.net/browse/SDK-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jferrao-itrbl:
Thanks for the notes.

1. Agreed — that pin note was stale. Both sides are 6.6.7; SDK-709 is the planned bump, not a current `pod install` failure. Description updated: dropped that paragraph and marked AC 11 done.

2. Kept the Android bridge no-op as specified (`wakeApp` is the inverse: Android-only real work). Native warning stays in logcat. Added a JS-layer `IterableLogger.log` on Android with the same message so the gap is visible in Metro when `logReactNativeSdkCalls` is on, then still call the native method. See `8c0841bf`.

rtlsilva (approved):

